PR #13, opened 2016-10-27
First cut at IDataSet interfaces based on Observables

mbudiu-vmw:


src/main/java/org/hiero/sketch/dataset/LocalDataSet.java
```diff
@@ -0,0 +1,47 @@
+package org.hiero.sketch.dataset;
```
lalithsuresh: Missing Javadocs for class

src/main/java/org/hiero/sketch/dataset/LocalDataSet.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public LocalDataSet(final T data) {
+        this.data = data;
+    }
+
+    @Override
+    public <S> Observable<PartialResult<IDataSet<S>>> map(final IMap<T, S> mapper) {
+        final PartialResult<S> initial = new PartialResult<S>(0.0, null);
+        final Observable<PartialResult<S>> start = Observable.just(initial);
+        final Observable<PartialResult<S>> trsf = mapper.apply(this.data);
```
lalithsuresh: Any particular reason you're referring to member variables with "this." all the time?

src/main/java/org/hiero/sketch/dataset/LocalDataSet.java
```diff
@@ -0,0 +1,47 @@
...
+        final Observable<PartialResult<S>> chain = start.concatWith(trsf);
+        final Observable<PartialResult<IDataSet<S>>> progress =
+                chain.map(e -> new PartialResult<IDataSet<S>>(e.deltaDone, null));
+        final PartialResultMonoid<S> mono = new PartialResultMonoid<S>(new NullMonoid<S>());
```
lalithsuresh: mono -> monoid

src/main/java/org/hiero/sketch/dataset/LocalDataSet.java
```diff
@@ -0,0 +1,47 @@
...
+        final Observable<PartialResult<S>> last = chain.reduce(mono.zero(), mono::add);
+        return progress.concatWith(last.map(e ->
+                new PartialResult<IDataSet<S>>(0.0, new LocalDataSet<S>(e.deltaValue))));
+    }
+
+    @Override
+    public <S> Observable<PartialResult<IDataSet<Pair<T, S>>>> zip(final IDataSet<S> other) {
+        if (!(other instanceof LocalDataSet<?>))
```
lalithsuresh: There are multiple instances of this throughout the patch, but please use braces for if statements even if its just a single statement underneath. Makes the code easier to read.

src/main/java/org/hiero/sketch/dataset/ParallelDataSet.java
```diff
@@ -0,0 +1,101 @@
...
+        this.children = children;
+    }
+
+    private int size() { return this.children.size(); }
+
+    @Override
+    public <S> Observable<PartialResult<IDataSet<S>>> map(final IMap<T, S> mapper) {
+        final ArrayList<Observable<Pair<Integer, PartialResult<IDataSet<S>>>>> obs =
+                new ArrayList<Observable<Pair<Integer, PartialResult<IDataSet<S>>>>>(this.size());
+        for (int i=0; i < this.size(); i++) {
```
lalithsuresh: Spacing "i = 0".

src/main/java/org/hiero/sketch/dataset/ParallelDataSet.java
```diff
@@ -0,0 +1,101 @@
...
+    public ParallelDataSet(final ArrayList<IDataSet<T>> children) {
+        this.children = children;
+    }
+
+    private int size() { return this.children.size(); }
+
+    @Override
+    public <S> Observable<PartialResult<IDataSet<S>>> map(final IMap<T, S> mapper) {
+        final ArrayList<Observable<Pair<Integer, PartialResult<IDataSet<S>>>>> obs =
+                new ArrayList<Observable<Pair<Integer, PartialResult<IDataSet<S>>>>>(this.size());
```
lalithsuresh: Again, why use this.<member/method> outside of constructors? Is there some ambiguity here?

src/main/java/org/hiero/sketch/dataset/ParallelDataSet.java
```diff
@@ -0,0 +1,101 @@
...
+            final Observable<Pair<Integer, PartialResult<IDataSet<S>>>> ci =
+                    this.children.get(i)
+                            .map(mapper)
+                            .map(e -> new Pair<Integer, PartialResult<IDataSet<S>>>(finalI, e));
+            obs.add(i, ci);
+        }
+        final Observable<Pair<Integer, PartialResult<IDataSet<S>>>> merged =
+                Observable.merge(obs);
+
+        final Observable<PartialResult<IDataSet<S>>> map = merged.filter(p -> p.second.deltaValue != null)
```
lalithsuresh: Name this variable something other than map. The method name is already 'map', and there's also several invocations of map on the child IDataSets and observables.

src/main/java/org/hiero/sketch/dataset/ParallelDataSet.java
```diff
@@ -0,0 +1,101 @@
...
+                .map(m -> new PartialResult<IDataSet<S>>(0.0, new ParallelDataSet<S>(m)));
+        final Observable<PartialResult<IDataSet<S>>> dones =
+                merged.map(p -> p.second.deltaDone / this.size())
+                        .map(e -> new PartialResult<IDataSet<S>>(e, null));
+        return dones.concatWith(map);
+    }
+
+    @Override
+    public <S> Observable<PartialResult<IDataSet<Pair<T, S>>>> zip(final IDataSet<S> other) {
+        if (!(other instanceof ParallelDataSet<?>))
```
lalithsuresh: More single statement ifs without braces.

src/main/java/org/hiero/sketch/dataset/ParallelDataSet.java
```diff
@@ -0,0 +1,101 @@
...
+        final int mySize = this.size();
+        if (mySize != os.size())
+            throw new RuntimeException("Different sizes for ParallelDatasets: " +
+                    mySize + " vs. " + os.size());
+        final ArrayList<Observable<Pair<Integer, PartialResult<IDataSet<Pair<T, S>>>>>> obs =
+                new ArrayList<Observable<Pair<Integer, PartialResult<IDataSet<Pair<T, S>>>>>>();
+        for (int i=0; i < mySize; i++) {
+            final IDataSet<S> oChild = os.children.get(i);
+            final IDataSet<T> tChild = this.children.get(i);
+            final Observable<PartialResult<IDataSet<Pair<T, S>>>> zip = tChild.zip(oChild).last();
```
lalithsuresh: This comment also applies in the LocalDataSet class.

Your zip method throws exceptions but they are unhandled here. Since you are using unchecked exceptions (RunTimeException), I guess your intent here is to bubble the exception all the way to the initial caller?

Typically, unchecked exceptions are used for things we do not expect the caller to recover from (like a divide by zero). However, it isn't clear that this is the case here. The map and zip functions come in at runtime against different datasets based on user inputs, which if invalid, should be correctly caught and reported back to the user. 

At some point, whatever is the client for such code should ideally have a "throws DatasetIncompatibleException" or so in my opinion. Since RuntimeException is unchecked, the client code is not forced to handle the exception.

src/main/java/org/hiero/sketch/dataset/ParallelDataSet.java
```diff
@@ -0,0 +1,101 @@
...
+            final int finalI = i;
+            obs.add(zip.map(
```
lalithsuresh: It's getting very difficult to discern the map operator of the observables from the map methods of IDataSet in the code. Can we rename the IDataSet map method here to something different?

src/main/java/org/hiero/sketch/dataset/ParallelDataSet.java
```diff
@@ -0,0 +1,101 @@
...
+        final Observable<PartialResult<IDataSet<Pair<T, S>>>> dones =
+                merged.map(p -> p.second.deltaDone / this.size())
+                        .map(e -> new PartialResult<IDataSet<Pair<T, S>>>(e, null));
+        return dones.concatWith(map);
+    }
+
+    @Override
+    public <R> Observable<PartialResult<R>> sketch(final ISketch<T, R> sketch) {
+        final ArrayList<Observable<PartialResult<R>>> obs = new ArrayList<Observable<PartialResult<R>>>();
+        final int mySize = this.size();
```
lalithsuresh: More use of "this" that isn't required.

src/main/java/org/hiero/sketch/dataset/api/PartialResult.java
```diff
@@ -0,0 +1,40 @@
...
+     */
+    public final T deltaValue;
+
+    /**
+     * Creates a partial result.
+     * @param deltaDone  How much more has been done.  A number between 0 and 1.
+     * @param deltaValue Extra result produced.
+     */
+    public PartialResult(double deltaDone, final T deltaValue) {
+        if (deltaDone < 0)
```
lalithsuresh: "deltaDone < 0" sounds like an exception or even something to check with an assert. Why assign deltaDone to 0 instead?

src/main/java/org/hiero/sketch/dataset/api/IProgressObserver.java
```diff
@@ -0,0 +1,9 @@
+package org.hiero.sketch.dataset.api;
```
lalithsuresh: Where is this used? Any implementations of the interface?

src/main/java/org/hiero/sketch/dataset/LocalDataSet.java
mbudiu-vmw: I have changed the settings to force braces for all multi-line ifs

src/main/java/org/hiero/sketch/dataset/LocalDataSet.java
mbudiu-vmw: Let's try with this style for a while and see what happens.

src/main/java/org/hiero/sketch/dataset/ParallelDataSet.java
mbudiu-vmw: The observables should take care of exceptions.

src/main/java/org/hiero/sketch/dataset/ParallelDataSet.java
mbudiu-vmw: It is the same method, really. 
I have changed the IMap object to contain an apply method instead of map.
But I hope to be able to eliminate the IMap object.

PR #15, opened 2016-10-30
Sample(k) and FastUtil

uwieder:
Implemented sample(k) and sample(k, seed). The implementation in PartialMembershipMapSparse is a placeholder that just returns the first k items the iterator provides. 
Also added the library FastUtil for the type specific sets we used. Even without tuning the tests run 5 times faster.

### DIFF
--- a/src/main/java/org/hiero/sketch/table/FullMembership.java
+++ b/src/main/java/org/hiero/sketch/table/FullMembership.java
@@ -3,6 +3,10 @@ package org.hiero.sketch.table;
 import org.hiero.sketch.table.api.IMembershipSet;
 import org.hiero.sketch.table.api.IRowIterator;
 
+import java.util.HashSet;
+import java.util.Random;
+import java.util.Set;
+
 /**
  * A IMembershipSet which contains all rows.
  */
@@ -30,6 +34,40 @@ public class FullMembership implements IMembershipSet {
     @Override
     public IRowIterator getIterator() {
         return new FullMemebershipIterator(this.rowCount);
+    }
+
+    /**
+     * Samples k items. Does not seed the random generator so generator is seeded using its default
+     * method. Sampled items are first placed in a Set. The procedure samples k times with
+     * replacement so it may be that the returned set has less than k distinct items
+     * @param k
+     * @return ImembershipSet instantiated as a Partial Sparse
+     */
+    @Override
+    public IMembershipSet sample(int k) {
+        Random randomGenerator = new Random();
+        return sampleUtil(randomGenerator, k);
+    }
+
+    /**
+     * Same as sample(k) but with the seed of the generator given as a parameter. The procedure
+     * samples k times with replacement so it return a set with less than k distinct items
+     * @param k
+     * @param seed
+     * @return ImembershipSet instantiated as a partial sparse
+     */
+    @Override
+    public IMembershipSet sample (int k, long seed) {
+        Random randomGenerator = new Random(seed);
+        return sampleUtil(randomGenerator, k);
+    }
+
+    //todo: use the best set implementation.
+    private IMembershipSet sampleUtil(Random randomGenerator, int k) {
+        Set S = new HashSet<Integer>();
+        for (int i=0; i < k; i++)
+            S.add(randomGenerator.nextInt(this.rowCount));
+        return new PartialMembershipSparse(S);
     }
 
     private static class FullMemebershipIterator implements IRowIterator {
--- a/src/main/java/org/hiero/sketch/table/FullMembership.java
+++ b/src/main/java/org/hiero/sketch/table/FullMembership.java
@@ -1,5 +1,6 @@
 package org.hiero.sketch.table;
 
+import it.unimi.dsi.fastutil.ints.IntOpenHashSet;
 import org.hiero.sketch.table.api.IMembershipSet;
 import org.hiero.sketch.table.api.IRowIterator;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/hiero/sketch/table/FullMembership.java
```diff
@@ -32,6 +36,40 @@ public IRowIterator getIterator() {
...
 
+    /**
+     * Samples k items. Does not seed the random generator so generator is seeded using its default
+     * method. Sampled items are first placed in a Set. The procedure samples k times with
+     * replacement so it may be that the returned set has less than k distinct items
+     * @param k
+     * @return ImembershipSet instantiated as a Partial Sparse
+     */
+    @Override
+    public IMembershipSet sample(int k) {
```
mbudiu-vmw: Are you sure this is the API we need? I would have expected that we provide a sampling rate rather than an expected number of samples. One benefit of that is that it does not depend on the set size and it can be used in a sketch.

src/main/java/org/hiero/sketch/table/PartialMembershipSparse.java
```diff
@@ -63,19 +91,53 @@ public int getSize() {
         return this.rowCount;
     }
 
+    /** TODO This is a PLACEHOLDER
```
mbudiu-vmw: It would be nice to remove this TODO.

src/test/java/org/hiero/sketch/MembershipTest.java
```diff
@@ -25,6 +26,19 @@ public void TestFullMembership() {
         assert(FM.isMember(7));
         assert(!FM.isMember(-2));
         assert(!FM.isMember(-10));
+        IMembershipSet mysample = FM.sample(5);
+        System.out.println("printing sample of 5 elements: ");
+        PrintMemebership(mysample);
+
+    }
+
+    private void PrintMemebership(IMembershipSet myset) {
```
mbudiu-vmw: The standard Java way is to have a toString() method. The benefit of this is that the debugger calls it for you automatically when you want to see the value. However, for large structures toString() should be done carefully: 
- use a StringBuilder, because string concatentation is linear time, so concatenating n strings turns into quadratic time
- perhaps truncate the result to a prefix if the data structure is too large (print an ellipsis)

src/test/java/org/hiero/sketch/MembershipTest.java
```diff
@@ -25,6 +26,19 @@ public void TestFullMembership() {
...
+        IRowIterator it = myset.getIterator();
+        int tmp = it.getNextRow();
+        while (tmp >=0 ) {
```
mbudiu-vmw: no space after 0

src/test/java/org/hiero/sketch/MembershipTest.java
```diff
@@ -35,13 +49,11 @@ public void TestPartialDense() {
         assert(PMD.isMember(6));
```
mbudiu-vmw: use assertEquals.
In fact I fixed this in my own pull request, but I haven't merged yet (and I have fixed other things in your tests). Take a look at my code.
In general, don't print in tests - it's useless when you have 1000 tests. Just use assertions.

src/main/java/org/hiero/sketch/table/PartialMembershipDense.java
```diff
@@ -65,6 +73,81 @@ public int getSize() {
     }
 
     @Override
+    public int getSize(boolean exact) {
```
mbudiu-vmw: Since getSize(true) is exactly the same semantically as getSize(), maybe you can instead have a getApproximateSize() method. Also, you may want getApproximateSize to return something else - perhaps a confidence in addition to the size?

src/main/java/org/hiero/sketch/table/PartialMembershipDense.java
```diff
@@ -65,6 +73,81 @@ public int getSize() {
...
+                tmprow = it.getNextRow();
+            }
+        }
+        return new PartialMembershipSparse(sampleSet);
+    }
+
+    /**
+     * Samples the base map for k items and then applies the filter on that set. Makes 10 attempts to reach k samples
+     * this way and then gives up and returns whatever was sampled.
+     * @param k
```
mbudiu-vmw: If you don't want to document the parameters please remove them from the comment.

src/main/java/org/hiero/sketch/table/api/IMembershipSet.java
```diff
@@ -15,6 +15,26 @@
...
+     * return the same sample set.
+     */
+    IMembershipSet sample(int k);
+
+    /**
+     * @return an ImembershipSet containing k samples from the membership map. The samples are made
+     * with replacement so may contain less than k distinct values. The pseudo-random generated
+     * is seeded with parameter seed, so subsequent calls with the same seed are guaranteed to return the same sample.
+     */
+    IMembershipSet sample(int k, long seed);
```
mbudiu-vmw: Please add here some binary set operations: intersection, union, difference and symmetric difference at least.

src/test/java/org/hiero/sketch/MembershipTest.java
```diff
@@ -25,6 +26,19 @@ public void TestFullMembership() {
         assert(FM.isMember(7));
         assert(!FM.isMember(-2));
         assert(!FM.isMember(-10));
+        IMembershipSet mysample = FM.sample(5);
```
mbudiu-vmw: Please use assertTrue and assertFalse.

src/main/java/org/hiero/sketch/table/PartialMembershipDense.java
uwieder: I actually like this interface, it does not clutter the API. I thing of getSize() as just setting the default to true (that's how I would write it in C++ where there are defaults). I don't see a reason to provide confidence intervals. We could have getAppSize(int k) where k is the number of samples used to drive the approximation, but really I don't see a real reason for this right now.

PR #18, opened 2016-11-02
Changes to ListColumn, row compartors.

parikg:
1. Changed structure of ListColumn classes (only use missing bitset when type is int/double)
2. Comparator now checks for missing values.
3. Added tests for ListColumn with missing values.
4. Test with a better comparison of TopK implementations.

### DIFF
--- a/src/test/java/org/hiero/sketch/StringListTest.java
+++ b/src/test/java/org/hiero/sketch/StringListTest.java
@@ -12,8 +12,6 @@ import static junit.framework.TestCase.*;
 class StringListTest {
 
     private final ColumnDescription desc = new ColumnDescription("test", ContentsKind.String, true);
-
-    /* Test for constructor using length and no arrays*/
     @Test
     public void testStringArrayZero() {
         final StringListColumn col = new StringListColumn(this.desc);
--- a/src/test/java/org/hiero/sketch/StringListTest.java
+++ b/src/test/java/org/hiero/sketch/StringListTest.java
@@ -1,0 +1,39 @@
+
+package org.hiero.sketch;
+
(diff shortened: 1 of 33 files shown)


src/test/java/org/hiero/sketch/StringListTest.java
```diff
@@ -0,0 +1,37 @@
...
+import static junit.framework.TestCase.*;
+
+/*
+ * Test for StringArrayColumn class.
+*/
+class StringListTest {
+
+    private final ColumnDescription desc = new ColumnDescription("test", ContentsKind.String, true);
+
+    /* Test for constructor using length and no arrays*/
```
mbudiu-vmw: Is this comment accurate? Maybe it should be deleted.

src/test/java/org/hiero/sketch/TreeVsHeapTest.java
```diff
@@ -0,0 +1,38 @@
...
+            for (int j = 1; j < this.inpSize; j++) {
+                this.randInp[j] = this.rn.nextInt(this.inpSize);
+            }
+
+            long startTime = System.nanoTime();
+            for (int j = 1; j < this.inpSize; j++) {
+                myHeap.push(this.randInp[j]);
+            }
+            long endTime = System.nanoTime();
+            System.out.format("Run %d: Heap takes %d ms, ",i, (endTime - startTime)/1000000);
```
mbudiu-vmw: Maybe you can use the new class for reporting perf?

parikg:
I fixed these and submitted new code. I think this pull request is still
open, so I dont need to submit a new request, right?

P.

PR #28, opened 2016-11-29
Merging Mihai's changes

uwieder:
Mihai's changes in the code broke some stuff. Completed the merge and submitting the pull request.


src/main/java/org/hiero/sketch/table/FullMembership.java
```diff
@@ -64,18 +62,58 @@ public IMembershipSet sample(final int k, final long seed) {
...
+        if (otherSet instanceof FullMembership)
+            return new FullMembership(Integer.min(this.rowCount, otherSet.getSize()));
+        return otherSet.intersection(this);
+    }
+
+    @Override
+    public IMembershipSet setMinus (final IMembershipSet otherSet) throws NullArgumentException {
+        if (otherSet == null)
+            throw new NullArgumentException("Can not perform setMinus with a null");
+        if (otherSet instanceof FullMembership) {
```
mbudiu-vmw: Is this case worth the effort? Won't the next case be almost just as fast?

src/main/java/org/hiero/sketch/table/FullMembership.java
```diff
@@ -64,18 +62,58 @@ public IMembershipSet sample(final int k, final long seed) {
...
+    }
+
+    @Override
+    public IMembershipSet copy() {
+        return new FullMembership(this.rowCount);
+    }
+
     private IMembershipSet sampleUtil(final Random randomGenerator, final int k) {
-        final IntOpenHashSet s = new IntOpenHashSet();
+        final IntSet S = new IntSet(k);
```
mbudiu-vmw: It's recommended to use lowercase names for variables, and uppercase for types.

src/main/java/org/hiero/sketch/table/HashUtil.java
```diff
@@ -0,0 +1,116 @@
+package org.hiero.sketch.table;
```
mbudiu-vmw: Maybe we should create a lib folder/package for this kind of stuff.
How about org.hiero.lib?

src/main/java/org/hiero/sketch/table/LazyMembership.java
```diff
@@ -127,17 +108,104 @@ public IMembershipSet sample(final int k) {
...
+            return 0;
+        int snumber = 0;
+        final IRowIterator it = sampleSet.getIterator();
+        int curr = it.getNextRow();
+        while (curr >= 0) {
+            if (this.filter.test(curr))
+                snumber++;
+            curr = it.getNextRow();
+        }
+        if (this.baseMap instanceof LazyMembership)
```
mbudiu-vmw: Are you assuming here that only LazyMembership has an expensive exactSize()?
If so you should probably document this at the very least.

src/main/java/org/hiero/sketch/table/LazyMembership.java
```diff
@@ -41,7 +40,7 @@ public LazyMembership(final IMembershipSet baseMap, final Predicate<Integer> fil
         if (filter == null) throw new NullArgumentException("PartialMembershipDense cannot be " +
                 "instantiated with a null filter");
         this.baseMap = baseMap;
-        this.rowCount = baseMap.getSize(false);
+        this.rowCount = 0;
```
mbudiu-vmw: I am wondering whether it won't be worthwhile to actually materialize a LazyMembership when you iterate over it.

src/main/java/org/hiero/sketch/table/SparseMembership.java
```diff
@@ -85,57 +91,116 @@ public int getSize() {
...
     }
 
     @Override
     public IRowIterator getIterator() {
-        return new SparseIterator(this.membershipMap);
+        return new SetSparseIterator(this.membershipMap);
+    }
+
+    @Override
+    public IMembershipSet union(final IMembershipSet otherSet) throws NullArgumentException{
```
mbudiu-vmw: Several space problems in this function.

src/main/java/org/hiero/sketch/table/SparseMembership.java
```diff
@@ -85,57 +91,116 @@ public int getSize() {
...
+    }
+    /**
+     * Estimates the size of a filter applied to an IMembershipSet
+     * @return an approximation of the size, based on a sample of size 20. May return 0.
+     * There are no strict guarantees on the quality of the approximation, but is good enough for
+     * initialization of a hash table sizes.
+     */
     private int estimateSize(final IMembershipSet baseMap, final Predicate<Integer> filter) {
-        final IMembershipSet sampleSet = baseMap.sample(sizeEstimationSampleSize);
+        final IMembershipSet sampleSet = baseMap.sample(20);
```
mbudiu-vmw: Please turn 20 into a static final value.

src/main/java/org/hiero/sketch/table/SparseMembership.java
```diff
@@ -144,24 +209,44 @@ private int estimateSize(final IMembershipSet baseMap, final Predicate<Integer>
...
-                return this.myIterator.next();
-            return -1;
+            if(!this.hasNext())
+                return -1;
+            --this.c;
+            if(this.mustReturnZero) {
+                this.mustReturnZero = false;
+                return 0;
+            }
+            final int[] key = this.membershipMap.key;
```
mbudiu-vmw: spaces

src/main/java/org/hiero/sketch/table/HashUtil.java
lalithsuresh: +1. I'd call it org.hiero.utils.

src/main/java/org/hiero/sketch/table/FullMembership.java
uwieder: No, this case does not call the isMember method. The isMember method may potentially use a predicate (in the Lazy case) and that might be more costly.

src/main/java/org/hiero/sketch/table/LazyMembership.java
uwieder: Currently, when we iterate over it for getSize() we just remember that we did it and don't do it again. I'm not sure it is worthwhile to instantiate a SparseMembership. The main difference is space. I'm thinking of a scenario where the baseMap is fullMembership and the predicate is fairly sparse. In this case LazyMembership will take O(1) space instead of O(n). There is always the possibility to convert it to a sparseMembership (via the sparseMembership constructor).

src/main/java/org/hiero/sketch/table/SparseMembership.java
uwieder: I don't see any spaces problem

src/main/java/org/hiero/sketch/table/LazyMembership.java
mbudiu-vmw: This is not about getSize, it's about the iterator itself.
There is the possibility, but the question is "who will do it".
The users of your classes only call these methods, all policies about data representation have to be part of your classes.

mbudiu-vmw:
Superseeded by #29

src/main/java/org/hiero/sketch/table/SparseMembership.java
mbudiu-vmw: There were several Exception{ and while (tmp >= 0_) cases which I fixed.

PR #30, opened 2016-12-01
code cleaning of indexing plus bucket

uwieder:
I made the two changes you requested plus modified the sample methods in sparseMembership in the nice way you did in lazyMembership. I also added a bucket class as preparation for histogram plus test for it. You can ignore the interface for histogram, it is an initial thought.

### DIFF
--- a/src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
+++ b/src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
@@ -10,8 +10,8 @@ public class Bucket<T> {
     private final boolean rightInclusive;
     private final Comparator<T> greater;
 
-    public Bucket(T leftBoundry, boolean leftInclusive,
-                       T rightBoundry, boolean rightInclusive, Comparator<T> greater) {
+    public Bucket(final T leftBoundry, final boolean leftInclusive,
+                  final T rightBoundry, final boolean rightInclusive, final Comparator<T> greater) {
         if (greater.compare(leftBoundry, rightBoundry) > 0)
             throw new IllegalArgumentException("Left boundry cannot be greater than right boundry");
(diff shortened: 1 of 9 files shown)


src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
```diff
@@ -0,0 +1,43 @@
+package org.hiero.sketch.spreadsheet;
+import java.util.*;
+/**
+ * A Bucket to be used as part of a histogram.
+ */
+public class Bucket<T> {
```
mbudiu-vmw: I think that this class is too general and not general enough at the same time.
- it's too general since it uses a generic T; all our datatypes are convertible to double, so probably we'll use doubles for all histograms
- it's not general enough since it does not support multi-dimensional histograms - there's just a left and a right boundary.
I propose to go the opposite way about this: instead of designing the most general histogram, let's build first a 1-D histogram of doubles with equally-spaced buckets, where the left is always inclusive except the rightmost bucket where the right is also inclusive. Then we build a 2-D histogram class, and then we see how we can generalize by merging the common code. So I would start by designing a OneDimensionalHistogram class, which maintains internally an array of buckets.

Please note that all histogram classes have to maintain a special bucket for "missing" data.

uwieder: I guess using doubles for all histograms is possible, though especially with a discrete domain may be somewhat awkward. I had each bucket have its own boundaries because I though we may want to cover the case buckets are not disjoint, but if that never happens design should be simpler. Multi-dimensional buckets are just 'AND' of one dimensional buckets. I would assume this is how we want to design them. We can discuss this further in a meeting.

PR #31, opened 2016-12-04
Changed the iterator as we discussed plus 1dhistogram

uwieder:
I began the Histogram1D. Take a look if this is what you were thinking of. If yes, I'll continue implementing a one dimensional histogram based on sampling. I'm thinking of writing a method  that continues to sample more and more data until it reaches the error bars specified by the user (a.la the paper we discussed a while ago).

### DIFF
--- a/src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
+++ b/src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
@@ -4,35 +4,60 @@ package org.hiero.sketch.spreadsheet;
  * A Bucket to be used as part of a histogram.
  */
 public class Bucket {
-    private final double leftBoundry;
-    private final double  rightBoundry;
+    private final double leftBoundary;
+    private final double  rightBoundary;
     private final boolean leftInclusive;
     private final boolean rightInclusive;
+    private Object minObject;
+    private Object maxObject;
+    private double minValue;
+    private double maxValue;
+    private int count;
 
-    public Bucket(final double leftBoundry, final boolean leftInclusive,
-                  final double rightBoundry, final boolean rightInclusive) {
-        if (leftBoundry > rightBoundry)
+    public Object getMinObject() { return this.minObject; }
+    public Object getMaxObject() { return this.maxObject; }
+    public double getMinValue() { return this.minValue; }
+    public double getMaxValue() { return this.maxValue; }
+
+    public Bucket(final double leftBoundary, final boolean leftInclusive,
+                  final double rightBoundary, final boolean rightInclusive) {
+        if (leftBoundary > rightBoundary)
             throw new IllegalArgumentException("Left boundry cannot be greater than right boundry");
-        else if ((leftBoundry == rightBoundry) && !(leftInclusive && rightInclusive))
+        else if ((leftBoundary == rightBoundary) && !(leftInclusive && rightInclusive))
             throw new IllegalArgumentException("Bucket defined over empty set");
-        this.leftBoundry = leftBoundry;
-        this.rightBoundry = rightBoundry;
+        this.leftBoundary = leftBoundary;
+        this.rightBoundary = rightBoundary;
         this.leftInclusive = leftInclusive;
         this.rightInclusive = rightInclusive;
     }
 
-    public Bucket(final double leftBoundry, final double rightBoundry) {
-         this(leftBoundry, true, rightBoundry, false);
+    public Bucket(final double leftBoundary, final double rightBoundary) {
+         this(leftBoundary, true, rightBoundary, false);
     }
 
     public boolean inBucket(final double item) {
 
-        if ((item < this.leftBoundry) || (item > this.rightBoundry))
+        if ((item < this.leftBoundary) || (item > this.rightBoundary))
             return false;
-        if (item == this.leftBoundry)
+        if (item == this.leftBoundary)
             return this.leftInclusive;
-        if (item == this.rightBoundry)
+        if (item == this.rightBoundary)
             return this.rightInclusive;
         return true;
+    }
+
+    //public void add(final double item, Object currObject) {
+    public void add(final double item ) {
+        this.count++;
+        if (item < this.minValue)
+        {
+            this.minValue = item;
+           // this.minObject = currObject;
+        }
+        if (item > this.maxValue)
+        {
+            this.maxValue = item;
+          //  this.maxObject = currObject;
+        }
     }
 }
--- a/src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
+++ b/src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
@@ -3,40 +3,35 @@ import java.util.*;
 /**
  * A Bucket to be used as part of a histogram.
  */
-public class Bucket<T> {
-    private final T leftBoundry;
-    private final T rightBoundry;
+public class Bucket {
+    private final double leftBoundry;
+    private final double  rightBoundry;
     private final boolean leftInclusive;
     private final boolean rightInclusive;
-    private final Comparator<T> greater;
 
-    public Bucket(final T leftBoundry, final boolean leftInclusive,
-                  final T rightBoundry, final boolean rightInclusive, final Comparator<T> greater) {
-        if (greater.compare(leftBoundry, rightBoundry) > 0)
+    public Bucket(final double leftBoundry, final boolean leftInclusive,
+                  final double rightBoundry, final boolean rightInclusive) {
+        if (leftBoundry > rightBoundry)
             throw new IllegalArgumentException("Left boundry cannot be greater than right boundry");
-        else if ((greater.compare(leftBoundry, rightBoundry) == 0 )
-                && !(leftInclusive && rightInclusive))
+        else if ((leftBoundry == rightBoundry) && !(leftInclusive && rightInclusive))
             throw new IllegalArgumentException("Bucket defined over empty set");
         this.leftBoundry = leftBoundry;
         this.rightBoundry = rightBoundry;
         this.leftInclusive = leftInclusive;
         this.rightInclusive = rightInclusive;
-        this.greater = greater;
     }
 
-    public Bucket(final T leftBoundry, final T rightBoundry, final Comparator<T> greater) {
-         this(leftBoundry, true, rightBoundry, false, greater);
+    public Bucket(final double leftBoundry, final double rightBoundry) {
+         this(leftBoundry, true, rightBoundry, false);
     }
 
-    public boolean inBucket(final T item) {
-        final int leftValue = this.greater.compare(item, this.leftBoundry);
-        final int rightValue = this.greater.compare(item, this.rightBoundry);
+    public boolean inBucket(final double item) {
 
-        if ((leftValue < 0) || (rightValue > 0))
+        if ((item < this.leftBoundry) || (item > this.rightBoundry))
             return false;
-        if (leftValue == 0)
+        if (item == this.leftBoundry)
             return this.leftInclusive;
-        if (rightValue == 0)
+        if (item == this.rightBoundry)
             return this.rightInclusive;
         return true;
     }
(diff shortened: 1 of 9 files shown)


src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
```diff
@@ -3,40 +3,35 @@
 /**
  * A Bucket to be used as part of a histogram.
  */
-public class Bucket<T> {
-    private final T leftBoundry;
-    private final T rightBoundry;
+public class Bucket {
+    private final double leftBoundry;
```
lalithsuresh: Boundry -> Boundary

src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
```diff
@@ -3,40 +3,35 @@
...
-    public Bucket(final T leftBoundry, final boolean leftInclusive,
-                  final T rightBoundry, final boolean rightInclusive, final Comparator<T> greater) {
-        if (greater.compare(leftBoundry, rightBoundry) > 0)
+    public Bucket(final double leftBoundry, final boolean leftInclusive,
+                  final double rightBoundry, final boolean rightInclusive) {
+        if (leftBoundry > rightBoundry)
             throw new IllegalArgumentException("Left boundry cannot be greater than right boundry");
-        else if ((greater.compare(leftBoundry, rightBoundry) == 0 )
-                && !(leftInclusive && rightInclusive))
+        else if ((leftBoundry == rightBoundry) && !(leftInclusive && rightInclusive))
```
lalithsuresh: If leftInclusive == true and rightInclusive == true, then it's okay for leftBoundary to be equal to the rightBoundary? That still is an empty set and should throw an error right?

src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
```diff
@@ -3,40 +3,35 @@
...
+         this(leftBoundry, true, rightBoundry, false);
     }
 
-    public boolean inBucket(final T item) {
-        final int leftValue = this.greater.compare(item, this.leftBoundry);
-        final int rightValue = this.greater.compare(item, this.rightBoundry);
+    public boolean inBucket(final double item) {
 
-        if ((leftValue < 0) || (rightValue > 0))
+        if ((item < this.leftBoundry) || (item > this.rightBoundry))
```
lalithsuresh: Doesn't look like this method will hit the later if statements below in some cases where it should.

Say,
leftBoundary = 10
rightBoundary = 15
leftInclusive = true
rightInclusive = false
item = 10

The if statement in line 30 will evaluate to false, and the method will therefore return false (when in fact with leftInclusive == true, the result should have been true).

These kind of cases should ideally come up in the unit tests. One way of catching them is to use the code coverage analysis from IntelliJ.

src/main/java/org/hiero/sketch/spreadsheet/Histogram1D.java
```diff
@@ -0,0 +1,88 @@
...
+public class Histogram1D implements IHistogram {
+    final private IColumn column;
+    final private IMembershipSet membershipSet;
+    final private double[] boundaries;
+    final private int[] buckets;
+    final private IStringConverter converter;
+    private int missingData;
+    private int outOfRange;
+
+    public Histogram1D(@NonNull IColumn column, @NonNull IMembershipSet membershipSet,
```
lalithsuresh: This seems like it will accept an empty "boundaries" array when it shouldn't. isSorted will return true and then line 25 will complain about negative array sizes.

src/main/java/org/hiero/sketch/spreadsheet/Histogram1D.java
```diff
@@ -0,0 +1,88 @@
...
+                placeInBucket(this.column.asDouble(currIndex, this.converter));
+            currIndex = myIter.getNextRow();
+        }
+    }
+
+    private void placeInBucket(final double entry) {
+        int index = Histogram1D.indexOf(this.boundaries, entry);
+        if (index >= 0)
+            this.buckets[index]++;
+        else this.outOfRange++;
```
lalithsuresh: Line break after else.

src/main/java/org/hiero/sketch/spreadsheet/Histogram1D.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    // Todo: Move to a base class at some point
+    public static boolean isSorted(final double [] a) {
+        for (int i = 0; i < (a.length - 1); i++) {
+            if (a[i] > a[i + 1]) {
```
lalithsuresh: This should be ">=" right? If there is an assumption somewhere that identical values cannot be in "a", that needs to be specified in the comments.

src/main/java/org/hiero/sketch/spreadsheet/Histogram1D.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * @param a array of sorted doubles, represent the buckets
+     * @param key
+     * @return index i such that a[i] <= key < a[i+1]. With the exception of the largest bucket
+     * in which case the right boundary is inclusive. If key < a[0] or key > a[length] returns -1
+     * Note that return is in [0..a.length-2].
+     * todo: Move to a base class at some point. Perhaps implement as binary search.
+     */
+    public static int indexOf(double[] a, double key) {
```
lalithsuresh: Definitely should be binary search. :)

src/main/java/org/hiero/sketch/spreadsheet/Histogram1D.java
```diff
@@ -0,0 +1,87 @@
...
+        final int index = Histogram1D.indexOf(this.boundaries, entry);
+        if (index >= 0)
+            this.buckets[index]++;
+        else this.outOfRange++;
+    }
+
+    // Todo: Move to a base class at some point
+    private static boolean isSorted(final double[] a) {
+        for (int i = 0; i < (a.length - 1); i++) {
+            if (a[i] > a[i + 1]) {
```
lalithsuresh: I think Github lost my comment.

Shouldn't this be ">="? If there's an assumption that "a" is sorted, then that needs to be documented.

uwieder: It's a bug that got fixed 'automatically', under the assumption that right inclusive is always false the array has to be strongly sorted. I'll document it.

src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
lalithsuresh: I see. I'll just add that these assumptions about expected behavior for the class are not obvious from the code, so it should be documented through comments.

src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
```diff
@@ -1,43 +1,63 @@
...
+    private final double leftBoundary;
+    private final double  rightBoundary;
     private final boolean leftInclusive;
     private final boolean rightInclusive;
-    private final Comparator<T> greater;
+    private Object minObject;
+    private Object maxObject;
+    private double minValue;
+    private double maxValue;
+    private int count;
```
mbudiu-vmw: Make this long; we need to support very large datasets.

src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
```diff
@@ -1,43 +1,63 @@
...
         return true;
     }
+
+    //public void add(final double item, Object currObject) {
+    public void add(final double item ) {
+        this.count++;
+        if (item < this.minValue)
+        {
+            this.minValue = item;
+           // this.minObject = currObject;
```
mbudiu-vmw: Why is this commented-out?

src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
```diff
@@ -1,43 +1,63 @@
...
+        }
+        if (item > this.maxValue)
```
mbudiu-vmw: You should have an 'else' here, you can save a few cycles.

src/main/java/org/hiero/sketch/spreadsheet/BucketSetDescription.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Contains the MetaData for the buckets held by an histogram
+ */
+public class BucketSetDescription {
+    public final int numOfBuckets;
+    public final boolean equalSizeBukets;
+    public final double minValue;
+    public final double maxValue;
+    public final double[] boundaries; //todo: Change the array to a unmodified list of some kind
```
mbudiu-vmw: You should have an abstract base class, and two derived classes: one for equal-sized buckets, and one for arbitrary buckets.
At least the derived classes should be called something like OneDBucketSet.

src/main/java/org/hiero/sketch/spreadsheet/BucketSetDescription.java
```diff
@@ -0,0 +1,71 @@
...
+        this.minValue = minValue;
+        this.maxValue = maxValue;
+        this.equalSizeBukets = true;
+        this.boundaries = new double[numOfBuckets + 1];
+        for (int i = 0; i < (numOfBuckets + 1); i++)
+            this.boundaries[i] = minValue + (((maxValue - minValue) * i) / numOfBuckets);
+    }
+
+    /** The assumption is that the all buckets are only left inclusive except the right one which
+     *  is right inclusive. Boundaries has to be strongly sorted.
```
mbudiu-vmw: the right one is both left and right inclusive.

src/main/java/org/hiero/sketch/spreadsheet/Histogram1D.java
```diff
@@ -0,0 +1,68 @@
...
+import org.hiero.sketch.table.api.IMembershipSet;
+import org.hiero.sketch.table.api.IRowIterator;
+import org.hiero.sketch.table.api.IStringConverter;
+
+/**
+ * One Dimensional histogram. Does not contain the column and membershipMap
+ * todo: work in progress
+ */
+public class Histogram1D implements IHistogram {
+   // final private IColumn column;
```
mbudiu-vmw: Please delete these comments.

src/main/java/org/hiero/sketch/spreadsheet/Histogram1D.java
```diff
@@ -0,0 +1,68 @@
...
+    private int outOfRange;
+    final private BucketSetDescription bucketDescription;
+
+    public Histogram1D(@NonNull final BucketSetDescription bucketDescription) {
+        this.bucketDescription = bucketDescription;
+        this.buckets = new Bucket[bucketDescription.numOfBuckets];
+        createBuckets();
+    }
+
+    private void createBuckets() {
```
mbudiu-vmw: Maybe this should be a method of the BucketSetDescription.

src/main/java/org/hiero/sketch/table/SparseMembership.java
```diff
@@ -193,39 +171,16 @@ private int estimateSize(@NonNull final IMembershipSet baseMap,
...
-            this.c = this.membershipMap.size();
-            this.mustReturnZero = membershipMap.getContainsZero();
-        }
+    private class SparseIterator implements IRowIterator {
+        final private IntSet.IntSetIterator mysetIterator;
 
-        public boolean hasNext() {
-            return this.c != 0;
+        private SparseIterator(@NonNull final IntSet mySet) {
+            this.mysetIterator = mySet.getIterataor();
```
mbudiu-vmw: typo in word Iterator

src/main/java/org/hiero/sketch/spreadsheet/Bucket.java
uwieder: I'll do the changes after the big merge. Don't wait for me.

PR #32, opened 2016-12-14
Histograms and Buckets

uwieder:
Left the sampling methods as a todo for next round. I'll do that once the basic design is stabilized.
Fixed a minor typo in the membership maps as well.


hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/Bucket1D.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public long getCount() { return this.count; }
+
+    public void add(final double item, @NonNull final Object currObject) {
+        if (this.count == 0) {
+            this.minValue = item;
+            this.minObject = currObject;
+            this.maxValue = item;
+            this.maxObject = currObject;
+        }
```
mbudiu-vmw: too many newlines: } else { on a single line

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/Bucket1D.java
```diff
@@ -0,0 +1,84 @@
...
+ * A one dimensional bucket designed for a Histogram1D
+ */
+public class Bucket1D {
+    private Object minObject;
+    private Object maxObject;
+    private double minValue;
+    private double maxValue;
+    private long count;
+
+    public Bucket1D() { }
```
mbudiu-vmw: Maybe this constructor could have two arguments: maxValue and minValue.
It should also initialize all fields.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/Bucket1D.java
```diff
@@ -0,0 +1,84 @@
...
+    public long getCount() { return this.count; }
+
+    public void add(final double item, @NonNull final Object currObject) {
+        if (this.count == 0) {
+            this.minValue = item;
+            this.minObject = currObject;
+            this.maxValue = item;
+            this.maxObject = currObject;
+        }
+        else if (item < this.minValue )
```
mbudiu-vmw: No space before closed parenthesis

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/Histogram1D.java
```diff
@@ -8,61 +8,73 @@
...
+        this.bucketDescription = bucketDescription;
+        this.buckets = new Bucket1D[bucketDescription.getNumOfBuckets()];
     }
 
     /**
-     * Creates the histogram explicitly and in full.
+     * Creates the histogram explicitly and in full. Should be called at most once.
      */
-    private void createHistogram(final IColumn column, final IMembershipSet membershipSet
+    public void createHistogram(final IColumn column, final IMembershipSet membershipSet
```
mbudiu-vmw: Space after comma, not before

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/IBucketsDescription1D.java
```diff
@@ -0,0 +1,34 @@
+package org.hiero.sketch.spreadsheet;
+
+/**
+ * meta data for one dimensional bucket set
```
mbudiu-vmw: I would make this comment more elaborate - saying that buckets are intended to represent a histogram.
Also, I would move the comment about which boundaries are inclusive here from the indexOf method.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/BucketsDescription1D.java
```diff
@@ -0,0 +1,111 @@
...
+            return this.numOfBuckets - 1;
+        int lo = 0;
+        int hi = this.boundaries.length - 1;
+        while (lo <= hi) {
+            int mid = lo + (hi - lo) / 2;
+            if (item < this.boundaries[mid]) hi = mid ;
+            else if (item >= this.boundaries[mid + 1]) lo = mid;
+            else return mid;
+        }
+        return -1;
```
mbudiu-vmw: Is this reachable? If not, throw an exception here.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/BucketsDescription1D.java
```diff
@@ -0,0 +1,111 @@
...
+        return this.getLeftBoundary(index + 1);
+    }
+
+    @Override
+    public int getNumOfBuckets() {
+        return this.numOfBuckets;
+    }
+
+    @Override
+    public double[] getBoundaries() {
```
mbudiu-vmw: Why make a copy?
You should add a comment that the caller is not supposed to mutate the boundaries.
But is this method even needed?

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/BucketsDescriptionEqSize.java
```diff
@@ -0,0 +1,79 @@
+package org.hiero.sketch.spreadsheet;
+
+/**
+ * MetaData for one dimensional buckets of equal size
+ */
+public class BucketsDescriptionEqSize implements IBucketsDescription1D {
+
```
mbudiu-vmw: Too many empty lines around here.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/BucketsDescriptionEqSize.java
```diff
@@ -0,0 +1,79 @@
...
+    private final double minValue;
+    private final double maxValue;
+    private final int numOfBuckets;
+
+
+    public BucketsDescriptionEqSize(final double minValue, final double maxValue, final int numOfBuckets) {
+        this.minValue = minValue;
+        this.maxValue = maxValue;
+        this.numOfBuckets = numOfBuckets;
```
mbudiu-vmw: Maybe you should check that maxValue >= minValue, just to be safe.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/BucketsDescriptionEqSize.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public int indexOf(final double  item) {
+        if (item < this.minValue || item > this.maxValue)
+            return -1;
+        if (item == this.maxValue)
```
mbudiu-vmw: Can you change this to item >= this.maxValue?
In principle doubles should never be compared for equality...

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/BucketsDescriptionEqSize.java
```diff
@@ -0,0 +1,79 @@
...
+
+
+    public BucketsDescriptionEqSize(final double minValue, final double maxValue, final int numOfBuckets) {
+        this.minValue = minValue;
+        this.maxValue = maxValue;
+        this.numOfBuckets = numOfBuckets;
+    }
+
+    @Override
+    public int indexOf(final double  item) {
```
mbudiu-vmw: You don't need two consecutive spaces.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/Bucket1D.java
```diff
@@ -0,0 +1,84 @@
...
+public class Bucket1D {
+    private Object minObject;
+    private Object maxObject;
+    private double minValue;
+    private double maxValue;
+    private long count;
+
+    public Bucket1D() { }
+
+    private Bucket1D(final long count, final double minValue,
```
mbudiu-vmw: This constructor can start by calling the other one:
{ this(minValue, maxValue); ... }

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/Histogram1D.java
```diff
@@ -8,61 +8,73 @@
...
                                 ,final IStringConverter converter ) {
+        if (this.buckets[0] != null) //a histogram had already been created
+            throw new IllegalAccessError("A histogram cannot be created twice");
+        for (int i = 0; i < this.bucketDescription.getNumOfBuckets(); i++)
+            this.buckets[i] = new Bucket1D();
```
mbudiu-vmw: The bucket does not have boundaries - you should probably use the new constructor I suggested.
I would move this code to the constructor.
In this way the constructor returns the "zero" for the monoid of histograms.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/Histogram1D.java
```diff
@@ -8,61 +8,73 @@
...
+     */
+    public Histogram1D union( @NonNull Histogram1D otherHistogram) {
+        if (!this.bucketDescription.equals(otherHistogram.bucketDescription))
+            throw new IllegalArgumentException("Histogram union without matching buckets");
+        Histogram1D unionH = new Histogram1D(this.bucketDescription);
+        for (int i = 0; i < unionH.bucketDescription.getNumOfBuckets(); i++)
+            unionH.buckets[i] = this.buckets[i].union(otherHistogram.buckets[i]);
+        unionH.missingData = this.missingData + otherHistogram.missingData;
+        unionH.outOfRange = this.outOfRange + otherHistogram.outOfRange;
+        return unionH;
```
mbudiu-vmw: You will have to think about addition of histograms on sampled data.
Probably the results have to be scaled up in order for addition to make sense.

hieroplatform/src/test/java/org/hiero/sketch/BucketTest.java
```diff
@@ -1,22 +1,28 @@
...
+        }
+        assertTrue(myBucket.getCount() == 100);
+        assertTrue(myBucket1.getCount() == 100);
+        assertTrue(myBucket1.getMinValue() == 0.5);
+        assertTrue(myBucket1.getMaxValue() == 99.5);
+        final Bucket1D myBucket2 = myBucket.union(myBucket1);
+        assertTrue(myBucket2.getCount() == 200);
+        assertTrue(myBucket2.getMinValue() == 0);
+        assertTrue(myBucket2.getMaxValue() == 99.5);
+        assertTrue(myBucket2.getMinObject().equals("0"));
```
mbudiu-vmw: You should use assertEquals instead of assertTrue.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/Bucket1D.java
uwieder: The maxValue and minValue are a function of the items placed in the bucket, not the boundary of the bucket. There is no reason to construct one with them. There is a private construction function that gets all the fields and is used in the union operation.
Java by default initializes all objects to null/zero so it's just a style issue, I added initializations for readability.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/Bucket1D.java
uwieder: See my previous comment. I'm not sure I understand why that's useful

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/Histogram1D.java
uwieder: The way the code is currently written, the individual buckets of the bucket array within Histogram1D are constructed only when a histogram is actually created (i.e. either by the createHistogram method or the union method). If they need to be initialized upon construction (for the zero histogram) I'll have to change the code here - for instance add a variable for the total number of items in the histogram (makes sense anyhow) and verify that it's zero. Either way, I don't see why I need to change the constructor for Bucket1D itself.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/BucketsDescription1D.java
uwieder: Indeed unreachable. Won't compile without a return statement though. I'll throw an exception instead.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/BucketsDescription1D.java
uwieder: I made a copy so that it is safe to pass on, if a comment is sufficient then I'll change it. It is not needed in any of the classes I wrote. I thought this would be useful for the front end when visualizing the histogram (instead of getting it bucket by bucket) . If not it could be deleted.

mbudiu-vmw:
Maybe the information has to be folded into the Bucket class – the buckets could have left and right margins.

PR #38, opened 2017-02-02
Changed the PSG to MersenneTwister from appache.commons.math. The PSG…

uwieder:
… is wrapped in a class called Randomness.
Also, I made all calls for randomness via a getInstance method.


hieroplatform/src/main/java/org/hiero/utils/Randomness.java
```diff
@@ -0,0 +1,34 @@
+package org.hiero.utils;
+
+import org.apache.commons.math3.random.MersenneTwister;
+
+public class Randomness {
```
mbudiu-vmw: This class should be documented carefully, since everyone who is using random number generators is expected to use this class. You should start by declaring an interface which has the last three methods in the signature. Then the class can extend the signature. In the README.md you should make it clear that people should never use java.util.Random, but instead this class.

Also, we should figure out the correct protocol to use this class. Who should initialize the seed? It could be initialized once, at application start-up time, or every time you make a call? Also, note that the hiero service is shared among many potential clients, so the random number generator will also be shared.

We should add a proper API for initializing.

Maybe the class should have only static methods? It's kind of weird to say:

```
Randomness i = Randomness.getInstance();
int x = r.getNext();
```

How about just

```
int x = Randomness.getNext();
```

hieroplatform/src/main/java/org/hiero/utils/Randomness.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public static Randomness getInstance() {
+        if (prg == null)
+            prg = new Randomness();
+        return prg;
+    }
+
+    public static Randomness getInstance(long seed) {
+        if (prg == null)
```
mbudiu-vmw: This function is very misleading, since it won't do anything if you have already called getInstance. You may be thinking you are setting the seed, but in fact you are not doing anything.

Let's separate retrieving the instance from intializing the seed; these should be two separate calls.

hieroplatform/src/main/java/org/hiero/utils/Randomness.java
mbudiu-vmw: Going to static methods is difficult to reverse; so let's stick with the old approach.
The method should be called createInstance().

mbudiu-vmw:
Can you please run a measurement test to see if this is slower than the traditional random?

hieroplatform/src/main/java/org/hiero/utils/Randomness.java
uwieder: static is hard because a PSG has a lot of state. It seems that this type of implementation is typically called getInstance(), but I don't mind changing the name. 
I'll document everything clearly and add an interface. Currently the class is initiated by the first call to getInstance. It is not clear to me who should initialize it. Does it really makes sense to separate the initialization from the 'createInstance()' (or getInstance()) methods?

hieroplatform/src/main/java/org/hiero/utils/Randomness.java
uwieder: Ok, this answers my previous questions...

uwieder:
Btw, I checked how scientific computing applications run monte-carlo simulations. They have several PRG's they use, most from the late 80's. The new packages actually use MersenneTwister so I'm confident it would be alright.

uwieder:
I ran performance tests. MersenneTwister is comparable to Random, and in if anything slightly faster. I found on stack exchange someone reporting that the same phenomena (i.e. MT being slightly faster than rand) occurs also in C++. Looking at the algo itself it makes sense.

hieroplatform/src/main/java/org/hiero/utils/Randomness.java
uwieder: It is not clear to me why write an interface. Do you imagine another IRandomness class in the project? An interface cannot include the 'getInstance' 'createInstance' functions. I think it makes sense (though not compulsory) that the PSG is a singleton. So the users will have to call specifically Randomness, and not refer to the Interface.

uwieder:
I made all the changes you requested. You can pull from master.

PR #42, opened 2017-02-10
Added a base class for 1DHisograms and a 1D sketch

uwieder:
Please take a look at the Sketch. I wrote for one class, I'll write the rest if this is good. I don't know how to test the create() method. I put a public method - getHistogram(..) similarly to getQuantile(..) in the quantile sketch.


hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/IHistogram1D.java
```diff
@@ -0,0 +1,22 @@
+package org.hiero.sketch.spreadsheet;
+
+import org.hiero.sketch.table.api.IColumn;
+import org.hiero.sketch.table.api.IMembershipSet;
+import org.hiero.sketch.table.api.IStringConverter;
+
+public interface IHistogram1D {
```
mbudiu-vmw: No empty lines.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/IHistogram1D.java
```diff
@@ -0,0 +1,22 @@
...
+
+    void createSampleHistogram(final IColumn column, final IMembershipSet membershipSet,
+                               final IStringConverter converter, double sampleRate, long seed);
+
+    void createHistogram(final IColumn column, final IMembershipSet membershipSet,
+                         final IStringConverter converter);
+
+    int getNumOfBuckets();
+
+    IHistogram1D union(IHistogram1D otherHistogram);
```
mbudiu-vmw: I would suggest removing this method from the interface - it causes problem with the typing.
Also, no empty line afterwards.

hieroplatform/src/test/java/org/hiero/sketch/HistSketchTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.hiero.sketch.spreadsheet.Histogram1DLight;
+import org.hiero.sketch.table.Table;
+import org.junit.Test;
+
+import static org.hiero.sketch.TableTest.getRepIntTable;
+import static org.junit.Assert.assertEquals;
+
+public class HistSketchTest {
+    @Test
+    public void Hist1DLightTest() {
```
mbudiu-vmw: How about a test where you apply this to an IDataSet<ITable>?

uwieder:
Added a test with IDataSet and it works nice. Cool! The only caveat is that I had to put some converter for the Int column (otherwise I get a null exception). I wrote a braindead converter, put it in the test folder and documented it as a class that would be scrapped once we have better converters.

hieroplatform/src/test/java/org/hiero/sketch/IntConverter.java
```diff
@@ -0,0 +1,13 @@
+package org.hiero.sketch;
```
mbudiu-vmw: This class should not be needed; it may signal a bug elsewhere.
What happens without it?

uwieder: I need to put some converter for the histogram. If I put a null instead I get a null pointer exception at ArrayColumn.asDouble(). It is triggered by this line:
double val = column.asDouble(currRow, converter);
in the createHistogram method of the histogram object.

uwieder: I don't think it's a bug, it is a reasonable behavoir  to demand that if the data type of the column is something other than double, then there must be an instantiation of the converter.

mbudiu-vmw: The converter is only needed for String columns, where it is mandatory.
For other data types we already have built-in converters.

PR #44, opened 2017-02-14
First GUI

mbudiu-vmw:
- simple sketch running all the way through GUI
- refactored ISketch and IMap interfaces to be simpler
- fixed observable tests to use TestSubscriber


hieroweb/web/src/hiero/web/InitialObject.java
```diff
@@ -2,34 +2,36 @@
...
-
     @HieroRpc
     void loadTable(@NonNull RpcRequest request, @NonNull Session session) {
         // TODO: look at request.  Now we just supply always the same table
-        String id = this.server.freshId();
         Table t = Table.testTable();
-        LocalDataSet<ITable> data = new LocalDataSet<ITable>(t);
-        TableTarget table = new TableTarget(id, data);
-        this.server.addObject(table);
+        final int parts = 5;
```
lalithsuresh: This still looks like test/demo code, but we don't have the plumbing to do the end-to-end test yet. I'd just mark this method with a TODO.

hieroweb/web/src/hiero/web/InitialObject.java
```diff
@@ -2,34 +2,36 @@
...
+        ArrayList<IDataSet<ITable>> fragments = new ArrayList<IDataSet<ITable>>();
+        for (int i = 0; i < parts; i++) {
+            LocalDataSet<ITable> data = new LocalDataSet<ITable>(t);
+            fragments.add(data);
+        }
+        IDataSet<ITable> big = new ParallelDataSet<ITable>(fragments);
+        TableTarget table = new TableTarget(big);
+        RpcObjectManager.instance.addObject(table);
```
lalithsuresh: Circular dependency with RpcObjectManager. See below.

hieroweb/web/src/hiero/web/RpcObjectManager.java
```diff
@@ -0,0 +1,70 @@
+package hiero.web;
+
+import com.google.gson.Gson;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+import java.util.HashMap;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+// Singleton pattern.
```
lalithsuresh: I'm not sure why we need the singleton pattern here. This is going to be a pain to mock and write tests for someday, especially when we have more than one web server talking to the Hiero backend.

There are two other classes that use RpcObjectManager: InitialObject and RpcServer. InitialObject and RpcObjectManager have a circular dependency, which should be fixed (RpcObjectManager uses addObject(initialObject)... and InitialObject.loadTable() later invokes RpcObjectManager.instance.addObject()). The logical order of dependencies looks like RpcServer -> RpcObjectManager -> InitialObject. And I'd suggest just using dependency injection to pass an RpcObjectManager instance to the RpcServer.

hieroweb/web/src/hiero/web/InitialObject.java
```diff
@@ -2,34 +2,36 @@
...
 
-        RpcReply reply = request.createReply(new RemoteTableId(id));
-        this.server.sendReply(reply, session);
+        RpcReply reply = request.createReply(table.idToJson());
```
lalithsuresh: It seems hacky to have the session management percolate all the way into the InitialObject. This method should just return a table at line 25 (and have some other class like RpcServer take that return value and process the remaining steps up to line 30).

hieroplatform/src/main/java/org/hiero/sketch/dataset/LocalDataSet.java
```diff
@@ -18,35 +20,46 @@ public LocalDataSet(final T data, final boolean separateThread) {
...
     }
 
     @Override
     public <S> Observable<PartialResult<IDataSet<Pair<T, S>>>> zip(final IDataSet<S> other) {
         if (!(other instanceof LocalDataSet<?>))
             throw new RuntimeException("Unexpected type in Zip " + other);
-        final LocalDataSet<S> lds = (LocalDataSet<S>) other;
-        final Pair<T, S> data = new Pair<T, S>(this.data, lds.data);
-        final LocalDataSet<Pair<T, S>> retval = new LocalDataSet<Pair<T, S>>(data);
+        LocalDataSet<S> lds = (LocalDataSet<S>) other;
```
mbudiu-vmw: I think we opt-in for fields, not for variables.
For variables final should be used when it clarifies the code, but I don't think we should enforce it.
It is very useful when you deal with multithreaded code and lambdas, which require effectively final variables.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/SummarySketch.java
```diff
@@ -31,7 +38,9 @@ public TableSummary() {
         public TableSummary add(TableSummary other) {
             Schema s = this.schema;
             if (this.schema == null)
```
mbudiu-vmw: An empty schema is different from a null schema.
Technically speaking, this is actually the OptionMonoid, for which null is the zero. I didn't use the OptionMonoid because it is used only for the schema, the rowCount is different. What happens here is that we have a pair of monoids combined: one for the schema OptionMonoid<Schema> and an addition for the rowcount.
So, yes, here the null is important. I will add a comment.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/SummarySketch.java
```diff
@@ -66,8 +75,13 @@ public TableSummary add(@NonNull TableSummary left, @NonNull TableSummary right)
 
     @NonNull
     @Override
-    public Observable<PartialResult<TableSummary>> create(@NonNull ITable data) {
+    public TableSummary create(@NonNull ITable data) {
+        try {
+            Thread.sleep(1000 * Randomness.getInstance().nextInt(5));
```
mbudiu-vmw: The sleep should not be here either, it was only used for testing the GUI. It will be removed.

hieroweb/web/src/hiero/web/InitialObject.java
mbudiu-vmw: Unfortunately this won't work, because a method may return multiple results. The server has no idea when the method is finished. So each @HieroRpc method is responsible for returning all replies and closing the session, which implies "end of replies".

It's even worse in general, because the replies may be generated on a separate thread. I will have to document this.

hieroweb/web/src/hiero/web/RpcObjectManager.java
mbudiu-vmw: The manager should be at the very bottom.
But the manager has to be initialized statically to hold the initial object. Not sure how to enforce this constraint in a different way.
Not familiar with dependency injection.

hieroplatform/src/main/java/org/hiero/sketch/spreadsheet/SummarySketch.java
lalithsuresh: Is it possible to have a Zero value in that case rather than null? You've used that style elsewhere and I think it is cleaner.

PR #60, opened 2017-04-04
Exact frequency computation + some more stuff

parikg:


hieroplatform/src/test/java/org/hiero/sketch/ExactFreqSketchTest.java
```diff
@@ -0,0 +1,57 @@
...
+                                .append(")").append("\n"));
+        return sb.toString();
+    }
+
+    @Test
+    public void EFSTest1() {
+        Table t = TestTables.testRepTable();
+        int maxSize = 10;
+        String s = getFrequencies(t, maxSize);
+        //System.out.println(s);
```
mbudiu-vmw: Would be nice to compare this to the correct value...

PR #66, opened 2017-04-20
Adding the missing functions for Randomness

uwieder:
Adding the wrapper for more functions in Randomness API

### DIFF
--- a/hieroplatform/src/main/java/org/hiero/table/FullMembership.java
+++ b/hieroplatform/src/main/java/org/hiero/table/FullMembership.java
@@ -66,7 +66,7 @@ public class FullMembership implements IMembershipSet {
     public IMembershipSet sample(final int k) {
         if (k >= this.rowCount)
             return new FullMembership(this.rowCount);
-        final Randomness randomGenerator = Randomness.getInstance();
+        final Randomness randomGenerator = new Randomness();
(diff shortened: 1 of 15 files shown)


uwieder:
By thread-safe do you mean remove the static getInstance method so that it is created normally via the constructor?

mbudiu-vmw:
Yes.
I think that if you pretty much remove the static it should work.

uwieder:
I don't think so. If I make it non static the getInstance() method cannot be called. I need to remove it completely and then change all the places where getInstance() is called with a standard constructor now?

mbudiu-vmw:
Or you can have getInstance just call a constructor and return the result.
But probably you are right.

uwieder:
I changed all the places where Randomness is called. **The NextKSketch test now fails!**
This is probably due to the fact that I commented out the getInstance(seed) calls. Parikshit needs to fix this, either by creating the instance of Randomness with a seed, or by using the .setSeed(seed) method.

mbudiu-vmw:
Why couldn't you use the constructor with an argument there?

uwieder:
Because Parikshit creates the instance in one method (with a seed) and then uses it in another method, relying on the fact it's a singleton. I can try to figure out where exactly to instantiate the seeds but I think it is better Parikshit do that.

PR #76, opened 2017-04-27
Murmur

uwieder:
Added HLogLog and HLogLog sketch for an approximate count of distinct elements. In the default setting the data structure uses 2^10 bytes and returns a count with an approximation of roughly 2%.
The data structure becomes inaccurate when the number of distinct items is very large, say at a million or so (I didn't thoroughly test where exactly). If we want good counts for this range of numbers we will need to work with 64 bits, starting with the hashcode.


hieroplatform/src/main/java/org/hiero/sketches/HLogLog.java
```diff
@@ -0,0 +1,111 @@
...
+    private final byte[] registers;
+    private final int seed;
+
+    /**
+     * @param logRegNum the logarithm of the number of registers. should be in 4...16
+     *        Going beyond 16 would make the data structure big for no good reason.
+     *        Setting logRegNum in at 10-12 should give roughly 2% accuracy most of the time
+     */
+    public HLogLog(int logRegNum, int seed) {
+        if (!spaceValid(logRegNum))
```
mbudiu-vmw: For static methods the convention is to use the class name: `HLogLog.spaceValid`

mbudiu-vmw: Maybe the spaceValid method could also throw the exception.
So then it is called checkSpaceValid, and it returns void.

hieroplatform/src/main/java/org/hiero/sketches/HLogLog.java
```diff
@@ -0,0 +1,111 @@
...
+     * adds the int 'value' to the data structure.
+     * Uses the first bits to identify the register and then counts trailing zeros
+     * @param value already assumed to be a random hash of the item
+     */
+    private void add(int value) {
+        int index =  value >>> (Integer.SIZE - this.logregNum);
+        byte zeros = (byte) (Integer.numberOfTrailingZeros(value) + 1);
+        if (zeros > this.registers[index])
+            this.registers[index] = zeros;
+
```
mbudiu-vmw: needless space.

hieroplatform/src/main/java/org/hiero/sketches/HLogLog.java
```diff
@@ -0,0 +1,111 @@
...
+        this.logregNum = logRegNum;
+        this.seed = seed;
+    }
+
+    /**
+     * adds the int 'value' to the data structure.
+     * Uses the first bits to identify the register and then counts trailing zeros
+     * @param value already assumed to be a random hash of the item
+     */
+    private void add(int value) {
```
mbudiu-vmw: If this does not work above 1M, then we should definitely use long instead.
Let's also call this itemHash instead of value.
I will file an issue for adding support for 64-bit hash codes.

hieroplatform/src/main/java/org/hiero/sketches/HLogLog.java
```diff
@@ -0,0 +1,111 @@
...
+        int index =  value >>> (Integer.SIZE - this.logregNum);
+        byte zeros = (byte) (Integer.numberOfTrailingZeros(value) + 1);
+        if (zeros > this.registers[index])
+            this.registers[index] = zeros;
+
+    }
+
+    /**
+     *
```
mbudiu-vmw: needless empty line

hieroplatform/src/main/java/org/hiero/sketches/HLogLog.java
```diff
@@ -0,0 +1,111 @@
...
+     * Creates a Hyperloglog data structure from a column and membership set. Uses the hash code
+     * of the objects in the column as identifier.
+     */
+    public void createHLL(IColumn column, IMembershipSet memSet) {
+        final IRowIterator myIter = memSet.getIterator();
+        int currRow = myIter.getNextRow();
+        while (currRow >= 0) {
+            int value = column.getObject(currRow).hashCode();
```
mbudiu-vmw: This will throw an exception if the object is missing.
Also, maybe you want to create a class HashCode which has a bunch of static methods for computing 64-bit hashcodes.
You may want to handle scalar specially, it you return an object it will perform boxing for every scalar, allocating an extra object.

hieroplatform/src/main/java/org/hiero/sketches/HLogLog.java
```diff
@@ -0,0 +1,111 @@
...
+            add(HashUtil.murmurHash3(this.seed, value));
```
mbudiu-vmw: this.add

hieroplatform/src/main/java/org/hiero/sketches/HLogLog.java
```diff
@@ -0,0 +1,111 @@
...
+        if ((otherHLL.regNum != this.regNum) || (otherHLL.seed != this.seed))
+            throw new IllegalArgumentException("attempted union of non matching HLogLog classes");
+        HLogLog result = new HLogLog(this.logregNum, this.seed);
+        for (int i = 0; i < this.regNum; i++)
+            result.registers[i] = (byte) Integer.max(this.registers[i], otherHLL.registers[i]);
+        return result;
+    }
+
+    /**
+     * @return an estimation of the number of distinct items
```
mbudiu-vmw: Why not call this distinctItemsEstimator?

hieroplatform/src/main/java/org/hiero/sketches/HLogLog.java
```diff
@@ -0,0 +1,111 @@
...
+            return Math.round(this.regNum * (Math.log(this.regNum / (double) zeroRegs)));
+        }
+    }
+
+    /**
+     * @param space the logarithm of the number of registers in the HLL structure.
+     * @return a boolean on whether it is within range. Going beyond 16 is too big and does not
+     * add accuracy. Going below 4 does not save space.
+     */
+    public static boolean spaceValid(int space) {
```
mbudiu-vmw: The argument should be called logSpaceSize

hieroplatform/src/main/java/org/hiero/sketches/HLogLogSketch.java
```diff
@@ -0,0 +1,46 @@
...
+import org.hiero.utils.Randomness;
+
+import javax.annotation.Nullable;
+
+public class HLogLogSketch implements ISketch<ITable, HLogLog> {
+    final String colName;
+    final int seed; //seed for the hash function of the HLogLog
+    /* the log of the #bytes used by each data structure. Should be in 4...16.
+     * More space means more accuracy. A space of 10-14 is recommended. */
+    final int spaceBound;
```
mbudiu-vmw: If this is the log, call it logSpaceSize.
The comment above should be converted to a Javadoc `/**` comment.

hieroplatform/src/main/java/org/hiero/sketches/HLogLogSketch.java
```diff
@@ -0,0 +1,46 @@
...
+    public HLogLogSketch(String colName) {
+        this.colName = colName;
+        this.seed = new Randomness().nextInt();
+        this.spaceBound = 12;
+    }
+
+    public HLogLogSketch(String colName, int spaceBound, int seed) {
+        this.colName = colName;
+        this.seed = seed;
+        if (!HLogLog.spaceValid(spaceBound))
```
mbudiu-vmw: This check seems duplicated, doing it once should be enough.

hieroplatform/src/test/java/org/hiero/sketch/HLLTest.java
```diff
@@ -0,0 +1,40 @@
...
+        final int size = 500000;
+        final int range = 10000;
+        final Randomness rn = new Randomness();
+        final IntArrayColumn col = IntArrayGenerator.getRandIntArray(size, range, "Test", rn);
+        final int accuracy = 12;
+        final int seed = rn.nextInt();
+        final HLogLog hll = new HLogLog(accuracy, seed);
+        final FullMembership memSet = new FullMembership(size);
+        hll.createHLL(col, memSet);
+        System.out.println("size is " + range + " approximat count: " + hll.appCount());
```
mbudiu-vmw: Can you convert this into an assertion somehow?
Perhaps by using some generous bounds on the returned value?
Also, you misspelled "approximate".

hieroplatform/src/test/java/org/hiero/sketch/HLLTest.java
```diff
@@ -0,0 +1,40 @@
...
+   // @Test
+    public void testHLLSketch() {
+        final int numCols = 1;
+        final int maxSize = 50;
+        final int bigSize = 100000;
+        final double rate = 1;
+        final SmallTable bigTable = TestTables.getIntTable(bigSize, numCols);
+        final String colName = bigTable.getSchema().getColumnNames().iterator().next();
+        final ParallelDataSet<ITable> all = TestTables.makeParallel(bigTable, bigSize / 10);
+        final HLogLog hll = all.blockingSketch(new HLogLogSketch(colName,12,1234567));
```
mbudiu-vmw: You are not doing anything with the result. At least check it is not null.

hieroplatform/src/main/java/org/hiero/sketches/HLogLog.java
uwieder: It will lose accuracy at around million. I've thought of some ways to (maybe) provide accurate estimates all the way up to a few million, but even if it works it will not go beyond that.

uwieder: I made changes that should increase accuracy in the large numbers. The accuracy would gradually deteriorate and provide under-estimates, possibly being useful up to a few tens of millions, its hard to tell without thorough testing.

uwieder:
I checked the scheme for larger numbers and it seems to run fine at 20,000,000. After that you start to see more often deviations of roughly 5-8%.

PR #106, opened 2017-06-16
Xx64hashing

uwieder:
Added support for xx hashing across all column types. The new dependency may perhaps make some of the methods in UtilHash redundant, but I left them there for now. HLL sketch now works with 64 bit hashes.


platform/src/main/java/org/hillview/table/ObjectArrayColumn.java
```diff
@@ -178,4 +179,9 @@ public static ObjectArrayColumn mergeColumns(final IColumn left, final IColumn r
         }
         return merged;
     }
+
+    @Override
+    public long hashCode64(int rowIndex, long seed) {
+        return LongHashFunction.xx(seed).hashLong((long) this.asDouble(rowIndex,null));
```
lalithsuresh: Is the seed going to be different on each invocation of hashCode64?

If not, you should not be invoking "LongHashFunction.xx(seed)" on every call to hashcode64() because it will allocate a new object every time you call it.

If in the lifetime of a column, the seed is going to be fixed, I would just instantiate one instance of LongHashFunction.xx(seed) in the constructor and re-use that instance for all subsequent calls to hashcode64() against that column.

This comment applies to all other uses of hashCode64 too!

platform/src/main/java/org/hillview/table/api/IColumn.java
```diff
@@ -116,4 +116,9 @@ default String getName() {
     }
 
     default ContentsKind getKind() { return this.getDescription().kind;}
+
+    /**
+     * @return A hash code for the item in the rowIndex. Exception if it is missing.
+     */
+    long hashCode64(int rowIndex, long seed);
```
mbudiu-vmw: For missing you should return a deterministic value, e.g., 0.

platform/src/main/java/org/hillview/table/ObjectArrayColumn.java
mbudiu-vmw: It looks wrong to convert the double to long, this loses information.
I think you want to use Double.doubleToRawLongBits.

platform/src/main/java/org/hillview/table/api/IColumn.java
mbudiu-vmw: I think that the API should not have a seed; the seed is part of the XX construction.
Unlike Randomness, it is probably OK to use a singleton static XX as the "global" XX function.
So you have a static method to set the seed at the very beginning. You also have another static method to get the XX, which replaces LongHashFunction.xx(seed).

platform/src/main/java/org/hillview/table/api/IDurationColumn.java
```diff
@@ -59,4 +62,12 @@ public int compare(final Integer i, final Integer j) {
             }
         };
     }
+
+    @Override
+    default long hashCode64(int rowIndex, long seed) {
+        if (isMissing(rowIndex))
+            throw new MissingException(this, rowIndex);
+        final Duration tmp = this.getDuration(rowIndex);
+        return LongHashFunction.xx(seed).hashLong(tmp.toMillis());
```
mbudiu-vmw: You the suitable Converter method to convert this to a double, and then apply the same algorithm as for doubles.

platform/src/main/java/org/hillview/table/api/IColumn.java
```diff
@@ -118,7 +118,12 @@ default String getName() {
     default ContentsKind getKind() { return this.getDescription().kind;}
 
     /**
-     * @return A hash code for the item in the rowIndex. Exception if it is missing.
+     * @return A 64 bit hash code for the item in the rowIndex.  xors the item with the seed and then runs xxhashing
```
mbudiu-vmw: I have another idea: how about replacing the seed with a reference to the XXHash?
The prototype is
```
long hashCode64(int rowIndex, XXHash hash);
```

uwieder: That's a good idea and a trivial change to make. I could do it in the future if we feel it is needed. One downside is that a reference to the XXHash would pass as part of the sketch, isn't there an overhead of serialization?

PR #107, opened 2017-06-16
Distributed Hillview: bring up server instances as processes and connect to them from the web server for data browsing

lalithsuresh:
1.  produces two runnable jars in platform/target on build. The first is platform-datacleaner-jar-with-dependencies.jar, that cleans up the data in ../data. The second is platform-jar-with-dependencies.jar that brings up HillviewServer instances.

2. Does the plumbing work in web/ to connect to backend servers over one or more RemoteDataSets.


README.md
```diff
@@ -65,11 +65,16 @@ $ mvn install
...
 ```
-$ mvn exec:java
-$ cd ..
+$ java -jar target/platform-datacleaner-jar-with-dependencies.jar
+```
+
+* Next, start an instance of a HillviewServer that listens on port 3569.
+
+```
+$ java -jar target/platform-jar-with-dependencies.jar 127.0.0.1:3569
```
mbudiu-vmw: Let's simplify the number of steps to run the demo as much as possible. Maybe we can just script the launch of the service and the web server.

platform/src/main/java/org/hillview/dataset/api/Empty.java
```diff
@@ -0,0 +1,14 @@
+package org.hillview.dataset.api;
+
+import java.io.Serializable;
+
+/**
+ * Represents a type used to bring up empty IDataSets.
+ */
+public class Empty implements Serializable {
```
mbudiu-vmw: Why does this implement Serializable?

platform/src/main/java/org/hillview/remoting/ClusterDescription.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public ClusterDescription(final List<HostAndPort> serverList) {
+        this.serverList = serverList;
+    }
+
+    public List<HostAndPort> getServerList() {
+        return this.serverList;
+    }
+
+
```
mbudiu-vmw: Maybe you can add a comment showing an example little json representation of a ClusterDescription.

web/src/main/java/org/hillview/InitialObjectTarget.java
```diff
@@ -30,18 +35,47 @@
...
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.stream.Stream;
 
 public class InitialObjectTarget extends RpcTarget {
+    private static final int HILLVIEW_DEFAULT_PORT = 3569;
```
mbudiu-vmw: Maybe these should be in a separate class HillviewService, part of the platform.

web/src/main/java/org/hillview/InitialObjectTarget.java
```diff
@@ -30,18 +35,47 @@
...
+        final ClusterDescription desc = new ClusterDescription(hostAndPorts);
+        this.initialize(desc);
+    }
+
+    private void initialize(final ClusterDescription description) {
+        final int numServers = description.getServerList().size();
+        assert numServers > 0;
+        if (numServers > 1) {
+            System.out.println("Creating PDS");
+            final ArrayList<IDataSet<Empty>> emptyDatasets = new ArrayList<>(numServers);
```
mbudiu-vmw: I don't like empty diamonds, although I agree the non-empty ones are redundant.

web/src/main/java/org/hillview/InitialObjectTarget.java
```diff
@@ -30,18 +35,47 @@
...
+        );
+        final ClusterDescription desc = new ClusterDescription(hostAndPorts);
+        this.initialize(desc);
+    }
+
+    private void initialize(final ClusterDescription description) {
+        final int numServers = description.getServerList().size();
+        assert numServers > 0;
+        if (numServers > 1) {
+            System.out.println("Creating PDS");
```
mbudiu-vmw: How about using the logging class?

web/src/main/java/org/hillview/InitialObjectTarget.java
```diff
@@ -30,18 +35,47 @@
...
+        final List<HostAndPort> hostAndPorts = Collections.singletonList(
+            HostAndPort.fromParts(LOCALHOST, HILLVIEW_DEFAULT_PORT)
+        );
+        final ClusterDescription desc = new ClusterDescription(hostAndPorts);
+        this.initialize(desc);
+    }
+
+    private void initialize(final ClusterDescription description) {
+        final int numServers = description.getServerList().size();
+        assert numServers > 0;
```
mbudiu-vmw: This should not be an assert; asserts are only for internal logic bugs, but this is input validation.
The cluster description could come from a user-supplied file.

PR #122, opened 2017-07-25
Add data range display

HanKruiger:


.gitignore
```diff
@@ -49,3 +49,35 @@ hs_err_pid*
 # data which is too big to put into git
 data/On_Time_On_Time*
 data/2016_*.csv
+data/readme.html
```
mbudiu-vmw: What is data/readme.html?

.gitignore
```diff
@@ -49,3 +49,35 @@ hs_err_pid*
...
+
+# Created by https://www.gitignore.io/api/macos
+
+### macOS ###
```
mbudiu-vmw: Do you really have all this junk on your disk?

web/src/main/webapp/package-lock.json
```diff
@@ -0,0 +1,511 @@
+{
```
mbudiu-vmw: What is this file? Is it needed?

web/src/main/webapp/package.json
```diff
@@ -14,8 +14,8 @@
     "url": "git://github.com/mbudiu-vmw/hiero.git"
   },
   "dependencies": {
-    "@types/d3": "^4.9.0",
-    "d3": "^4.0.5",
+    "@types/d3": "^4.10.0",
```
mbudiu-vmw: Do we have to manually upgrade, or will webpack cause these packages to be upgraded automatically?

Are you using any functionality which is new in d3 version 10?

web/src/main/webapp/table.ts
```diff
@@ -772,12 +772,38 @@ export class TableView extends RemoteObject
             return "?";  // TODO
     }
 
+    public drawDataRange(canvas: HTMLCanvasElement, position: number, count: number) : void {
```
mbudiu-vmw: I don't think this method has any reason to be public.

web/src/main/webapp/table.ts
```diff
@@ -772,12 +772,38 @@ export class TableView extends RemoteObject
...
+        var ctx = canvas.getContext('2d');
+
+        // A little hack to make canvas look prettier on retina displays.
```
mbudiu-vmw: So far we don't have any target-specific code.
Ideally this would be done with CSS instead of straight code.

web/src/main/webapp/table.ts
```diff
@@ -772,12 +772,38 @@ export class TableView extends RemoteObject
...
+        // (Source: https://coderwall.com/p/vmkk6a/how-to-make-the-canvas-not-look-like-crap-on-retina)
+        ctx.scale(2, 2);
+        canvas.style.width = '100px';
+        canvas.style.height = '20px';
+        canvas.width = 2 * 100;
+        canvas.height = 2 * 20;
+
+        let x = canvas.width * position / this.rowCount;
```
mbudiu-vmw: You may need to adjust x if width is 1 - otherwise the rectangle may spill out.

web/src/main/webapp/table.ts
```diff
@@ -772,12 +772,38 @@ export class TableView extends RemoteObject
...
+        let width = Math.max(1, canvas.width * count / this.rowCount);
+        ctx.fillStyle = 'rgb(200, 200, 200)';
+        ctx.fillRect(x, 0, width, canvas.height);
+
+        ctx.textBaseline = "middle";
```
mbudiu-vmw: Please use css for this; there is a hillview.css file, where you can add your style definitions.
Just use a clear class name for this canvas.

web/src/main/webapp/table.ts
```diff
@@ -772,12 +772,38 @@ export class TableView extends RemoteObject
...
+
     public addRow(row : RowView, cds: ColumnDescription[]) : void {
         let trow = this.tBody.insertRow();
 
+        let position = this.startPosition + this.dataRowsDisplayed;
+
         let cell = trow.insertCell(0);
-        cell.style.textAlign = "right";
-        cell.textContent = significantDigits(this.startPosition + this.dataRowsDisplayed);
+        let canvas = document.createElement("canvas");
```
mbudiu-vmw: I think you should actually create the canvas inside the drawDataRange function; in this way, how you draw it is entirely hidden in that function.

.gitignore
HanKruiger: Mostly the .DS_Store files. But I figured I'd add other known files as well. I'll change it to just *.DS_Store.

.gitignore
HanKruiger: It's a description of the dataset. It's included in the .zip that is downloaded by `data/download.sh`.

web/src/main/webapp/package.json
HanKruiger: I'm not sure about what webpack will upgrade automatically, but this was the current version in npm when I configured the GUI.

I'm not using any new functionality, so we might as well use older versions. But I don't see any problems with using the newer version either.

.gitignore
mbudiu-vmw: Then maybe you should modify download.sh to remove the readme.html as well.

web/src/main/webapp/table.ts
HanKruiger: Unfortunately, the things that are drawn *on* the canvas are not HTML elements, so their styles cannot be governed by CSS... Only the properties of the canvas itself (basically, just its dimensions) can be.
Using just HTML objects for the visualization would impose a lot of restrictions, I think. Perhaps for this case it's doable, but I reckon we really need a canvas for the more interesting visualizations.

HanKruiger: Oh, I completely forgot about d3. I'll check out what I can do with that.

web/src/main/webapp/package-lock.json
HanKruiger: As far as I understand, it makes sure that everyone building the front-end uses the same package versions. It is advised that this file is checked in to VCS. From the [npm docs](https://docs.npmjs.com/files/package-lock.json):
> package-lock.json is automatically generated for any operations where npm modifies either the node_modules tree, or package.json. It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.

HanKruiger:
I've fixed the width and contrast issues we've spoken about. Tested it in Chrome, Safari & Firefox, and don't see any issues.
I also made the script rename the readme.
Can you merge this, or are there necessary changes for this feature?

I'll use separate branches for separate features/issues from now on.

Issue #123, opened 2017-07-25
Websocket address is hard coded


lalithsuresh:
rpc.ts has the following line:

const HillviewServiceUrl : string = "ws://localhost:8080";

We need to determine the URL dynamically instead.

PR #124, opened 2017-07-25
Fix for issue #123

mbudiu-vmw:


mbudiu-vmw:
@lalithsuresh , please merge this if you like it.

HanKruiger:
This breaks the web sockets for me. On my system, `window.location.href` is `"http://localhost:8080/"`, and I get the following error message:
```
SyntaxError: Failed to construct 'WebSocket': The URL's scheme must be either 'ws' or 'wss'. 'http' is not allowed.
```
So perhaps the `http` prefix has to be changed to `ws`?

mbudiu-vmw:
We should have fixed this already; please try the latest version in master.

PR #132, opened 2017-07-31
Equality filter map

HanKruiger:
It works for Integers and Strings now.

Let me know if I should do something differently. (Even silly things like code formatting.)

### DIFF
--- a/platform/src/main/java/org/hillview/table/EqualityFilter.java
+++ b/platform/src/main/java/org/hillview/table/EqualityFilter.java
@@ -1,9 +1,9 @@
 package org.hillview.table;
 
-import com.sun.tools.javac.util.Assert;
 import org.hillview.table.api.IColumn;
 import org.hillview.table.api.ITable;
 import org.hillview.utils.Converters;
+import org.junit.Assert;
 
 import javax.annotation.Nullable;
 import java.util.Objects;
@@ -29,10 +29,13 @@ public class EqualityFilter implements TableFilter {
         // Check the types. Just Strings and Integers for now.
         switch (column.getDescription().kind) {
             case Category:
-                Assert.check(compareValue instanceof String);
+            case String:
+            case Json:
+            case Date:
+                Assert.assertTrue(compareValue instanceof String);
                 break;
             case Integer:
-                Assert.check(compareValue instanceof Integer);
+                Assert.assertTrue(compareValue instanceof Integer);
                 break;
         }
     }
--- a/platform/src/main/java/org/hillview/table/EqualityFilter.java
+++ b/platform/src/main/java/org/hillview/table/EqualityFilter.java
@@ -1,19 +1,22 @@
 package org.hillview.table;
 
+import com.sun.tools.javac.util.Assert;
 import org.hillview.table.api.IColumn;
 import org.hillview.table.api.ITable;
 import org.hillview.utils.Converters;
 
 import javax.annotation.Nonnull;
-
+import javax.annotation.Nullable;
+import java.util.Objects;
 
 /**
  * This filter maps a given Table to a Table that only contains the given value in the specified column.
  */
 public class EqualityFilter implements TableFilter {
     @Nonnull
-    private String columnName;
-    private Object compareValue;
+    private final String columnName;
+    private final Object compareValue;
+    @Nullable
     private IColumn column;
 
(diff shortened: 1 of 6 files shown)


platform/src/main/java/org/hillview/table/EqualityFilter.java
```diff
@@ -0,0 +1,52 @@
...
+import org.hillview.table.api.IColumn;
+import org.hillview.table.api.ITable;
+import org.hillview.utils.Converters;
+
+/**
+ * This filter maps a given Table to a Table that only contains the given value in the specified column.
+ * @param <T>
+ */
+public class EqualityFilter<T> implements TableFilter {
+    private String columnName;
```
HanKruiger: I used the `@Nonnull` annotation from `javax.annotations`. It that the one you meant?

platform/src/main/java/org/hillview/table/EqualityFilter.java
```diff
@@ -0,0 +1,58 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Objects;
+
+/**
+ * This filter maps a given Table to a Table that only contains the given value in the specified column.
+ */
+public class EqualityFilter implements TableFilter {
+    @Nonnull
```
mbudiu-vmw: My mistake, this is not needed; by default everything is non-null.
Only nullable is needed.

platform/src/main/java/org/hillview/table/EqualityFilter.java
```diff
@@ -0,0 +1,58 @@
...
+        this.compareValue = value;
+    }
+
+    @Override
+    public void setTable(ITable table) {
+        this.column = table.getColumn(this.columnName);
+
+        // Check the types. Just Strings and Integers for now.
+        switch (column.getDescription().kind) {
+            case Category:
```
mbudiu-vmw: This case also works for String and JSON. So just add case String: etc.

platform/src/main/java/org/hillview/utils/TestTables.java
```diff
@@ -18,6 +18,7 @@
 
 package org.hillview.utils;
 
+import com.sun.tools.javac.util.Assert;
```
mbudiu-vmw: You should use the unit test assert; please look at the other test files to see which import they use.

platform/src/main/java/org/hillview/utils/TestTables.java
```diff
@@ -65,6 +64,45 @@ public static Table testRepTable() {
     }
 
     /**
+     * Can be used for testing large tables with strings.
+     * @param size Number of rows in the table
+     * @param others Array of options in the "Name" column
+     * @param count Number of occurrences of the 'test' string.
+     * @param test The
```
mbudiu-vmw: incomplete comment.
Do not feel the need to document things which are obvious. But these are good comments.

platform/src/main/java/org/hillview/utils/TestTables.java
```diff
@@ -65,6 +64,45 @@ public static Table testRepTable() {
...
+     */
+    public static Table testLargeStringTable(int size, String[] others, int count, String test) {
+        ColumnDescription c0 = new ColumnDescription("Name", ContentsKind.Category, false);
+        ColumnDescription c1 = new ColumnDescription("Age", ContentsKind.Integer, false);
+
+        Assert.check(!Arrays.asList(others).contains(test));
+        Random random = new Random();
+        ArrayList<String> names = new ArrayList<String>();
+        ArrayList<Integer> ages = new ArrayList<Integer>();
+        for (int i = 0; i < size - count; i++){
```
mbudiu-vmw: Space between ) and {

platform/src/main/java/org/hillview/utils/TestTables.java
HanKruiger: The issue was that the junit assert wasn't in the `main` scope, only in the `test` scope. I've added it as a dependency.

mbudiu-vmw: Then why did all the other tests work?

HanKruiger: Because they're in the `test` folder, and `TestTables` is (contrary to what the name suggests) not in the `test` folder.

PR #133, opened 2017-08-01
Remove iml files from git

lalithsuresh:
This prevents the IntelliJ iml files from being tracked by git. They're being constructed based on the maven pom.xml files anyway.


mbudiu-vmw:
Let's try an see whether anything breaks.

PR #139, opened 2017-08-07
Linear projection map

HanKruiger:
- I've added [JBlas](http://jblas.org/) as a dependency for the linear algebra computations.


platform/src/main/java/org/hillview/maps/LinearProjectionMap.java
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * This map takes a list of column names and a projection matrix,
+ * and applies the projection matrix to the matrix that is
+ * constructed from the table by horizontally stacking the
+ * specified columns. The resulting table is a copy of the old
+ * table, with the additional projected columns added to it.
+ */
+public class LinearProjectionMap implements IMap<ITable, ITable> {
+
```
mbudiu-vmw: no need for this empty line

platform/src/main/java/org/hillview/maps/LinearProjectionMap.java
```diff
@@ -0,0 +1,72 @@
...
+    private final DoubleMatrix projectionMatrix;
+    private final String[] colNames;
```
mbudiu-vmw: You should probably document the projection matrix - in particular, the order of columns has to be the same as the colNames array.

platform/src/main/java/org/hillview/maps/LinearProjectionMap.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    public ITable apply(ITable data) {
+        // The data matrix that has the observations as rows
+        DoubleMatrix mat = data.getNumericMatrix(this.colNames, null);
+
+        // The projection along the directions in this.projectionMatrix.
+        // The projected rows are rows in this matrix too.
+        DoubleMatrix proj = mat.mmul(this.projectionMatrix.transpose());
```
mbudiu-vmw: Does BLAS have a method to multiply directly with a transpose?

platform/src/main/java/org/hillview/maps/LinearProjectionMap.java
```diff
@@ -0,0 +1,72 @@
...
+
+        this.projectionMatrix = projectionMatrix;
+        this.colNames = colNames;
+        this.numLowDims = projectionMatrix.rows;
+    }
+
+    @Override
+    public ITable apply(ITable data) {
+        // The data matrix that has the observations as rows
+        DoubleMatrix mat = data.getNumericMatrix(this.colNames, null);
```
mbudiu-vmw: I suspect this could be very expensive; it entails copying the entire data for all these columns. Moreover, this data is only temporarily used during the multiplication and then becomes garbage.

The best way to do this would be to have the actual data implement some matrix interface. Is there an IDoubleMatrix interface? Then you just implement get(int i, int j) for this interface, without copying the data.

platform/src/main/java/org/hillview/maps/LinearProjectionMap.java
```diff
@@ -0,0 +1,72 @@
...
+        List<IColumn> columns = new ArrayList<IColumn>();
+        Iterable<IColumn> inputColumns = data.getColumns();
+        for (IColumn inputColumn : inputColumns) {
+            columns.add(inputColumn);
+        }
+
+        // Add all the projections to the columns.
+        for (int j = 0; j < this.numLowDims; j++) {
+            String colName = String.format("LinearProjection%d", j);
+            ColumnDescription colDesc = new ColumnDescription(colName, ContentsKind.Double, false);
```
mbudiu-vmw: This should probably have "true" at the end: if your original membershipsets have gaps then the result will have missing values too.

platform/src/main/java/org/hillview/maps/LinearProjectionMap.java
```diff
@@ -0,0 +1,72 @@
...
+            DoubleArrayColumn column = new DoubleArrayColumn(colDesc, data.getMembershipSet().getMax());
```
mbudiu-vmw: One other way we can implement this is by creating a DoubleSparseColumn. We should seriously consider this.

platform/src/main/java/org/hillview/table/api/ITable.java
```diff
@@ -77,4 +78,48 @@
...
+
+    /**
+     * Returns a DoubleMatrix that is a column vector corresponding to the given column name.
+     * This copies the entire column into the new matrix.
+     *
+     * @param colName Name of the column that is desired.
+     * @param converter String converter to use for string data.
+     * @return The matrix with the numeric data from the specified column.
+     */
+    default DoubleMatrix getNumericColumn(String colName, IStringConverter converter) {
```
mbudiu-vmw: This is very expensive. Ideally we should never need this.

platform/src/main/java/org/hillview/utils/TestTables.java
```diff
@@ -183,6 +186,32 @@ public static SmallTable getHeavyIntTable(final int numCols, final int size, fin
...
+    }
+
+    /**
+     * Assumes that all columns are numeric, and returns the matrix corresponding to the data in the table.
+     * @param table The table that we want to convert.
+     * @return The DoubleMatrix with the same data as in the table.
+     */
+    @SuppressWarnings("ConstantConditions")
+    public static DoubleMatrix toDoubleMatrix(ITable table) {
+        String[] colNames = table.getSchema().getColumnNames().toArray(new String[]{});
```
mbudiu-vmw: Is the argument to toArray - the empt string array - needed?

platform/src/test/java/org/hillview/sketch/MatrixTest.java
```diff
@@ -0,0 +1,75 @@
...
+    @Test
+    public void testSimpleMatrix() {
+        // Setting elements
+        DoubleMatrix mat = new DoubleMatrix(2, 2);
+        mat.put(0, 0, 2.3);
+        mat.put(1, 0, 42.42);
+        mat.put(0, 1, 3.14);
+        mat.put(1, 1, 3.91992);
+        Assert.assertEquals(mat.get(0, 0), 2.3, Math.ulp(2.3));
+        Assert.assertEquals(mat.get(1, 0), 42.42, Math.ulp(42.42));
```
mbudiu-vmw: Interesting, I didn't know about ulp

platform/src/main/java/org/hillview/table/api/ITable.java
HanKruiger: I'll remove it from ITable for now. Perhaps I'll add the functionality later (elsewhere) when we need JBLAS for the nontrivial linear algebra.

PR #140, opened 2017-08-08
Implemented two-view operations (for tables)

mbudiu-vmw:


mbudiu-vmw:
I will amend this PR with updates that make it work on all views, not just tables.
Unfortunately I have moved some code around, so the diff will look larger than it is.
Perhaps there won't be a need for you to review it again.

PR #141, opened 2017-08-08
Full correlation matrix sketch and eigenvector computation

HanKruiger:
The full correlation sketch uses BasicColStats (that are the results of previous sketches). I'm not particularly sure if this is the right way to do it.

Most of it is based on how SampleCorrelationSketch did it, but this sketch uses all rows, and uses the BasicColStats to scale the columns, as it should for the correlation matrix.

In addition to the correlation matrix, I've also added a piece of code that uses JBLAS to compute the eigenvectors of a (symmetric) matrix. I tested this (in unison with the correlation matrix sketch) in PCATest to compute principal components, and it looks like it is working. (I.e., the test succeeds, and it makes sense.)


platform/src/main/java/org/hillview/maps/LinearProjectionMap.java
```diff
@@ -0,0 +1,67 @@
...
+import org.hillview.table.api.*;
+import org.jblas.DoubleMatrix;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This map takes a list of column names and a projection matrix and applies the projection matrix to the matrix that is
+ * constructed from the table by horizontally stacking the specified columns. The resulting table is a copy of the old
+ * table, with the additional projected columns added to it. The new columns are named 'LinearProjection{i}'.
```
mbudiu-vmw: I think that the names of the new columns should be members; with this algorithm you cannot run the LinearProjection twice on the same data.

platform/src/main/java/org/hillview/maps/LinearProjectionMap.java
```diff
@@ -0,0 +1,67 @@
...
+     * The projection matrix is structured as follows: Every row is a vector
+     * that is projected on. The ordering of the columns is the same as the order
+     * of the column names in colNames.
+     */
+    private final DoubleMatrix projectionMatrix;
+    private final String[] colNames;
+    private final int numProjections;
+    private final IStringConverter converter;
+
+    public LinearProjectionMap(String[] colNames, DoubleMatrix projectionMatrix, IStringConverter converter) {
```
mbudiu-vmw: The converter is Nullable.

platform/src/main/java/org/hillview/maps/LinearProjectionMap.java
```diff
@@ -0,0 +1,67 @@
...
+        Iterable<IColumn> inputColumns = table.getColumns();
+        for (IColumn inputColumn : inputColumns) {
+            columns.add(inputColumn);
+        }
+
+        // Compute and add all the projections to the columns.
+        for (int j = 0; j < this.numProjections; j++) {
+            String colName = String.format("LinearProjection%d", j);
+            ColumnDescription colDesc = new ColumnDescription(colName, ContentsKind.Double, true);
+            DoubleArrayColumn column = new DoubleArrayColumn(colDesc, table.getMembershipSet().getMax());
```
mbudiu-vmw: Please add here a // TODO: create and use a SparseColumn

platform/src/main/java/org/hillview/sketches/CorrMatrix.java
```diff
@@ -42,6 +42,17 @@ public void update(int i, int j, double val) {
         this.rawMatrix[i][j] += val;
     }
 
+    /**
+     * Sets a new value at (i, j) that is a weighted sum of this matrix's value and the other one, weighted by the
```
mbudiu-vmw: There is no "other one". Do you mean "val"?

platform/src/main/java/org/hillview/sketches/FullCorrelationSketch.java
```diff
@@ -0,0 +1,81 @@
...
+import org.hillview.table.api.*;
+import org.hillview.utils.Converters;
+
+import javax.annotation.Nullable;
+import java.security.InvalidParameterException;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This class is very similar to the SampleCorrelationSketch, apart from two important differences. First, it uses
```
mbudiu-vmw: This comment should be kept, but please add a comment that describes what this is really doing.
The user should not need to browse the other class to understand this one. Maybe you can just steal some of the comments from the other class.

platform/src/main/java/org/hillview/sketches/FullCorrelationSketch.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+        IColumn[] iCols = new IColumn[this.colNames.size()];
+        for (int l=0; l < this.colNames.size(); l++)
+            iCols[l] = table.getColumn(this.colNames.get(l));
+        CorrMatrix cm = new CorrMatrix(this.colNames);
+        IRowIterator rowIt = table.getRowIterator();
+        int i = rowIt.getNextRow();
+        double valJ, valK;
+        while (i != -1) {
+            for (int j = 0; j < this.colNames.size(); j++) {
```
mbudiu-vmw: Maybe you can swap the order of these two loops?

platform/src/main/java/org/hillview/sketches/FullCorrelationSketch.java
```diff
@@ -0,0 +1,81 @@
...
+                valJ = iCols[j].asDouble(i, null);
+                // Remove the mean to center the data in the column.
+                valJ -= this.basicColStatsMap.get(this.colNames.get(j)).getMoment(1);
```
mbudiu-vmw: If you swap the order of the loops, then you can move this code out of the inner loop. There are several dereferences here for each element access.

platform/src/main/java/org/hillview/sketches/FullCorrelationSketch.java
```diff
@@ -0,0 +1,81 @@
...
+                        "integer or double: " + col);
+        }
+        IColumn[] iCols = new IColumn[this.colNames.size()];
+        for (int l=0; l < this.colNames.size(); l++)
+            iCols[l] = table.getColumn(this.colNames.get(l));
+        CorrMatrix cm = new CorrMatrix(this.colNames);
+        IRowIterator rowIt = table.getRowIterator();
+        int i = rowIt.getNextRow();
+        double valJ, valK;
+        while (i != -1) {
```
mbudiu-vmw: Let's do i >= 0

platform/src/main/java/org/hillview/sketches/FullCorrelationSketch.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Nullable
+    @Override
+    public CorrMatrix add(@Nullable CorrMatrix left, @Nullable CorrMatrix right) {
+        left = Converters.checkNull(left);
+        right = Converters.checkNull(right);
+        for (int i = 0; i < this.colNames.size(); i++)
+            for (int j = i; j < this.colNames.size(); j++)
+                left.updateWeighted(i, j, right.get(i, j), right.count);
```
mbudiu-vmw: In general we prefer not to update in place if possible - it is much easier to achieve thread-safety if data is immutable.
Since the correlation matrix is not very large it is OK to allocate a new one for the result.

platform/src/main/java/org/hillview/utils/BlasConversions.java
```diff
@@ -0,0 +1,57 @@
...
+import org.hillview.table.api.*;
+import org.jblas.DoubleMatrix;
+import org.jblas.ranges.AllRange;
+import org.jblas.ranges.PointRange;
+
+import java.util.Arrays;
+
+/**
+ * This class contains methods that convert from/to Tables to/from DoubleMatrices. These methods copy all data, so they
+ * should only be used as a last resort, and preferably only when it is known that the Table is reasonably small.
```
mbudiu-vmw: I would remove "last resort" and keep the thing about small tables.

platform/src/main/java/org/hillview/utils/BlasConversions.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public class BlasConversions {
+    /**
+     * Convert from an ITable to a DoubleMatrix. This copies all data from the table, so it should only be used as a
+     * last resort.
+     * @param table Table that is to be converted.
+     * @param colNames Names of the table that have to be converted.
```
mbudiu-vmw: Names of the columns

platform/src/main/java/org/hillview/utils/BlasConversions.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param converter String converter for converting string data to numeric data.
+     * @return DoubleMatrix with the table's columns interpreted as doubles.
+     */
+    public static DoubleMatrix toDoubleMatrix(ITable table, String[] colNames, IStringConverter converter) {
```
mbudiu-vmw: The string converter is Nullable.

platform/src/test/java/org/hillview/sketch/CorrelationTest.java
```diff
@@ -0,0 +1,71 @@
...
+            BasicColStatSketch statSketch = new BasicColStatSketch(col.getName(), null);
+            BasicColStats bcs = statSketch.create(table);
+            bcsMap.put(col.getName(), bcs);
+        }
+
+        FullCorrelationSketch fcs = new FullCorrelationSketch(colNames, bcsMap);
+        CorrMatrix cm = fcs.create(table);
+
+        DoubleMatrix corrMatrix = new DoubleMatrix(cm.getCorrelationMatrix());
+        DoubleMatrix eigenVectors = LinAlg.eigenVectors(corrMatrix, 2);
```
mbudiu-vmw: If you know what the output should be then you should also check it. Since the input matrix is constant, it is fine to inline here the output values.

platform/src/main/java/org/hillview/sketches/FullCorrelationSketch.java
HanKruiger: That is not really possible, since we keep track of how many rows were processed to calculate the mean on the fly (in the new changed I pushed). And keeping track of every column individually would be a bit inconvenient.

platform/src/main/java/org/hillview/maps/LinearProjectionMap.java
HanKruiger: I've added an extra argument to the constructor for the base name of the new columns. The names are now "{baseName}{i}".

PR #146, opened 2017-08-16
PCA frontend and additional datasets

HanKruiger:
- PCA is now accessible from the frontend, but it only works for numeric data that has no missing values.
- The dataset I've added are:
  - MNIST: Handwritten digit images. The columns are the pixels in the image, the rows are images. The value in the column signifies the greyscale value of the pixel.
    - It is split in two parts: 
    - The download script triggers a Python script to convert the downloaded data to csv. This is not ideal, since it's an additional dependency, but I did not have a quick way of converting it otherwise. Ideally we do this with Java as well.
    - Doing PCA on this dataset takes about 11 seconds, which is quite slow. Python libraries I use do the same in about 1-2 seconds.
  - Segmentation: Every row in this dataset contains a set of features extracted from a 3x3 patch of an image. They're manually labeled by classes like 'sky', 'grass', 'concrete', etc. The features (columns) are things like the average hue or saturation, and number of edges going through the patch.
    - Since this dataset has fewer columns (just 19), PCA is a lot faster on this.
    - These kinds of datasets show how visualization can be used to assist in designing computer vision systems. It helps the designer to find relevant image features.


platform/src/main/java/org/hillview/sketches/FullCorrelationSketch.java
```diff
@@ -15,6 +18,7 @@
  * the BasicColStats's.
  */
 public class FullCorrelationSketch implements ISketch<ITable, CorrMatrix> {
+    private static final Logger LOG = Logger.getLogger(FullCorrelationSketch.class.getName());
```
mbudiu-vmw: I believe that this class will need to be modified since the mean has to be computed across all parts of a dataset.
Talk to Parikshit if it is not clear how to do this in one pass.
The actual correlation matrix will be computed on the root node only; this sketch will compute enough information to enable the computation.

platform/src/main/java/org/hillview/storage/CsvFileReader.java
```diff
@@ -111,6 +111,7 @@ public ITable read() throws IOException {
             settings.setIgnoreTrailingWhitespaces(true);
             settings.setEmptyValue("");
             settings.setNullValue(null);
+            settings.setMaxColumns(1024);
```
mbudiu-vmw: If you have a schema file use the number of columns from the schema.

web/src/main/webapp/table.ts
```diff
@@ -216,14 +217,21 @@ export class TableView extends RemoteObject
...
+                    { text: "Refresh", action: () => { this.refresh(); } },
+                    { text: "All columns", action: () => { this.showAllRows(); } },
+                    { text: "No columns", action: () => { this.setOrder(new RecordOrder([])); } }
+                ])
+            },
             {
                 text: "Combine", subMenu: combineMenu(this)
+            },
+            {
+                text: "Operation", subMenu: new TopSubMenu([
```
mbudiu-vmw: This should probably move to the context menu and should apply to the selected columns.

web/src/main/webapp/table.ts
```diff
@@ -716,6 +725,46 @@ export class TableView extends RemoteObject
...
+                valid = false;
+                message += "\n  * Column '" + colName  + "' is not numeric.";
+            }
+            if (allowMissing) {
+                valid = false;
+                message += "\n  * Column '" + colName + "' has missing values.";
+            }
+        });
+
+        if (valid) {
```
mbudiu-vmw: You should also check that you have at least 3 columns.

HanKruiger:
It works now, by using the same technique as the heavy hitters do for chaining the operations. So it's ready for review once again.

HanKruiger:
I've rebased it to the current master branch!

PR #149, opened 2017-08-18
Add equality filter by cell value

HanKruiger:


web/src/main/webapp/table.ts
```diff
@@ -870,9 +879,31 @@ export class TableView extends RemoteObject
...
+                            e.preventDefault();
+                            if (this.contextMenu != null) {
+                                this.contextMenu.remove();
+                            }
+                            this.contextMenu = new ContextMenu([
+                                {text: "Filter for " + cellValue, action: () => this.equalityFilter(cd.name, cellValue) },
+                                {text: "Filter for not " + cellValue, action: () => this.equalityFilter(cd.name, cellValue, true) }
+                            ]);
+
+                            document.body.appendChild(this.contextMenu.getHTMLRepresentation());
```
mbudiu-vmw: Can you please use the show() method of the context menu instead of appending the representation directly?

HanKruiger: Fixed this! There's now a single context menu (per TableView) that is now hidden/shown instead of removed/created.

PR #165, opened 2017-08-28
Heat map arrays with indicator overlays

HanKruiger:


HanKruiger:
This can be merged now, if it's up to me. I still need to add the color/axis legend to this, but this can be done quickly later.

web/src/main/webapp/rpc.ts
```diff
@@ -333,3 +333,32 @@ export class ZipReceiver extends Renderer<string> {
         rr.invoke(rec);
     }
 }
+
+// This collector waits for an operation to finish and calls a callback when
+// the operation has completed.
+export class CallbackCollector<T> extends Renderer<T> {
```
mbudiu-vmw: I like this.
Maybe it's better to make this a base class which is extended by various renderers.
Let's call it something like OnCompleteRenderer: a renderer which only executes when everything is completed.

web/src/main/webapp/table.ts
```diff
@@ -41,6 +44,14 @@ import {Dialog} from "./dialog";
 // return an enum from a string.
 export type ContentsKind = "Category" | "Json" | "String" | "Integer" | "Double" | "Date" | "Interval";
 
+export function isNumeric(kind: ContentsKind): boolean {
+    return kind == "Integer" || kind == "Double";
+}
+
+export function isCategorical(kind: ContentsKind): boolean {
+    return kind == "Category" || kind == "String"
```
mbudiu-vmw: This is wrong, strings are not categories.
Maybe you can call this isString or something else. But I suspect that you don't really want this function.

web/src/main/webapp/table.ts
```diff
@@ -765,7 +778,19 @@ export class TableView extends RemoteObjectView
         }
     }
 
+    private heatMapArray(): void {
+        let selectedColumns: string[] = [];
+        this.selectedColumns.forEach((col) => selectedColumns.push(col));
```
mbudiu-vmw: apparently there are simpler ways to clone an array in javascript.
Maybe you want to add a util.ts function that clones an array?

web/src/main/webapp/heatMapArray/types.ts
```diff
@@ -0,0 +1,17 @@
+import {IColumnDescription} from "../table"
```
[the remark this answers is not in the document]
mbudiu-vmw: Please note that you have to add some js files to .gitignore when you create new .ts files.
Do you really need a separate folder for this? It only seems to have 2 files.

web/src/main/webapp/heatMapArray/view.ts
```diff
@@ -0,0 +1,288 @@
...
+import {RangeInfo, TableView, Schema, IColumnDescription, isNumeric, isCategorical, DistinctStrings} from "../table";
+import {BasicColStats} from "../histogramBase";
+import {FullPage, Size, Resolution, IHtmlElement, ScrollBar, IScrollTarget} from "../ui";
+import {Pair, Triple} from "../util";
+import {ColumnAndRange} from "../histogramBase";
+import {HeatMapData} from "../heatmap";
+import {HeatMapArrayArgs, HeatMapArrayData} from "./types";
+import d3 = require('d3');
+
+class ColorMap {
```
mbudiu-vmw: Maybe we can reuse this for the 2d heatmap too. Do you want to refactor that code to use it?

web/src/main/webapp/heatMapArray/view.ts
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    private map(x: number) {
+        return d3.interpolateWarm(x);
+    }
+}
+
+class CompactHeatMapView implements IHtmlElement {
+    // Size of a single heatmap in pixels
+    public static readonly size: Size = {
```
mbudiu-vmw: I am worried about this being hardwired.
First of all, maybe you can reuse the Resolution class.
Second, you should try to compute the size; e.g., if there are only 3 categories, you should make each heat-map 1/3 of the page width instead of hardwiring the size here.

web/src/main/webapp/heatMapArray/view.ts
```diff
@@ -0,0 +1,288 @@
...
+        }
+        this.topLevel.appendChild(document.createElement("p")).textContent = binLabel;
+        this.chart = d3.select(this.topLevel).append("svg")
+            .attr("width", CompactHeatMapView.size.width)
+            .attr("height", CompactHeatMapView.size.height)
+    }
+
+    public put(x, y, val) {
+        this.chart.append("rect")
+            .attr("x", x * Resolution.minDotSize)
```
mbudiu-vmw: You cannot hardwire the dotsize, it has to be computed as the width divided by the number of buckets.
This changes when you resize the window.

web/src/main/webapp/heatMapArray/view.ts
```diff
@@ -0,0 +1,288 @@
...
+        height: 200
+    }
+    private static maxTextLabelLength = 10;
+
+    private topLevel: HTMLElement;
+    private chart: any;
+
+    constructor(private binLabel: string) {
+        this.topLevel = document.createElement("div")
+        if (binLabel.length > CompactHeatMapView.maxTextLabelLength) {
```
mbudiu-vmw: Please add a utility function to truncate text.

web/src/main/webapp/heatMapArray/view.ts
```diff
@@ -0,0 +1,288 @@
...
+    private offset: number;
+
+    constructor(remoteObjectId: string, page: FullPage, args: HeatMapArrayArgs) {
+        super(remoteObjectId, page);
+        this.args = args;
+        this.offset = 0;
+        if (isNumeric(args.cds[0].kind) && isNumeric(args.cds[1].kind) && isCategorical(args.cds[2].kind))
+            this.initiateNumNumCat();
+        else if (isNumeric(args.cds[0].kind) && isNumeric(args.cds[1].kind) && isNumeric(args.cds[2].kind))
+            this.initiateNumNumNum();
```
mbudiu-vmw: else throw

web/src/main/webapp/heatMapArray/view.ts
```diff
@@ -0,0 +1,288 @@
...
+        // Compute the number of heat maps that fits on the page.
+        let canvasSize = Resolution.getCanvasSize(this.getPage());
+        // +2 to account for 1 pixel of border on each side.
+        let numCols = Math.floor((canvasSize.width - ScrollBar.barWidth) / (CompactHeatMapView.size.width + 2));
+        let numRows = Math.floor(canvasSize.height / (CompactHeatMapView.size.height + 2));
+        return numCols * numRows;
+    }
+
+    public refresh() {
+
```
mbudiu-vmw: Please implement this function; it is needed when the browser page is resized.

web/src/main/webapp/heatMapArray/view.ts
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    public scrolledTo(position: number): void {
+        let oldOffset = this.offset;
+        this.offset = Math.min(
+            Math.floor(position * this.args.uniqueStrings.length),
+            this.args.uniqueStrings.length - this.maxNumHeatMaps()
```
mbudiu-vmw: Please create a function totalHeatMaps which just returns args.uniqueStrings.length.

web/src/main/webapp/heatMapArray/view.ts
```diff
@@ -0,0 +1,288 @@
...
+            this.offset + this.maxNumHeatMaps(),
+            this.args.uniqueStrings.length - this.maxNumHeatMaps()
+        );
+        console.log(`Page down: ${oldOffset} --> ${this.offset}`);
+        this.updateZBins();
+        this.initiateHeatMaps();
+    }
+
+    public updateView(heatMapsArray: HeatMapArrayData) {
+        // Clean up before updating.
```
mbudiu-vmw: Can you check whether data == null first?

web/src/main/webapp/heatMapArray/view.ts
```diff
@@ -0,0 +1,288 @@
...
+        this.heatMaps.forEach((heatMap) => {
+            heatMap.setColors(colorMap);
+        });
+        this.scrollBar.setPosition(this.offset / this.args.uniqueStrings.length, (this.offset + this.args.zBins.length) / this.args.uniqueStrings.length);
+    }
+
+    // Entry point. Start the appropriate sketches for the given arguments.
+    public start(args: HeatMapArrayArgs) {
+        this.args = args;
+        if (isNumeric(args.cds[0].kind) && isNumeric(args.cds[1].kind) && isCategorical(args.cds[2].kind))
```
mbudiu-vmw: This code is duplicated. Can you factor it out?

web/src/main/webapp/heatmap3d.ts
```diff
@@ -0,0 +1,804 @@
...
+class HeatMapLegend implements IHtmlElement {
+    private outer: HTMLElement;
+
+    // Fields for the axes.
+    private static axesTicks = 3;
+    private static readonly axesSize: Size = {
+        width: 200,
+        height: 200
+    };
+    private static readonly axesMargin = {
```
mbudiu-vmw: Can you instead reuse Resolution? (Assuming that the margins can be the same as there).

web/src/main/webapp/heatmap3d.ts
```diff
@@ -0,0 +1,804 @@
...
+        top: 40, right: 20, bottom: 40, left: 55
+    };
+
+    private axesG: any;
```
mbudiu-vmw: For the `any` fields maybe you can add a comment indicating what they are supposed to be.

web/src/main/webapp/heatmap3d.ts
```diff
@@ -0,0 +1,804 @@
...
+    private xScale: ScaleLinear<number, number> | ScaleTime<number, number>;
+    private yScale: ScaleLinear<number, number> | ScaleTime<number, number>;
+    private marker: any;
+    private xLine: any;
+    private yLine: any;
+    private xText: any;
+    private yText: any;
+
+    // Fields for color legend
+    private static readonly colorLegendSize: Size = {
```
mbudiu-vmw: Can you reuse Resolution?

web/src/main/webapp/heatmap3d.ts
```diff
@@ -0,0 +1,804 @@
...
+            .attr("stroke", "white")
+            .attr("visibility", "hidden")
+        this.valIndicatorText = this.colorLegendG.append("text")
+            .attr("text-anchor", "middle")
+            .attr("alignment-baseline", "baseline")
+            .attr("x", HeatMapLegend.colorLegendSize.width / 2)
+            .attr("y", HeatMapLegend.colorLegendSize.height + HeatMapLegend.colorLegendMargin.bottom - 5);
+    }
+
+    public update(mouseX, mouseY, val) {
```
mbudiu-vmw: please add types to the arguments.

web/src/main/webapp/heatmap3d.ts
```diff
@@ -0,0 +1,804 @@
...
+                .attr("x2", scaledVal)
+                .attr("visibility", "visible");
+            this.valIndicatorText.text(val);
+        } else {
+            this.valIndicatorLine.attr("visibility", "hidden")
+            this.valIndicatorText.text("0");
+        }
+    }
+
+    static invert(v: number, scale: AnyScale, kind: ContentsKind, buckets: string[]): string {
```
mbudiu-vmw: This code seems copy-and-paste, maybe we can refactor it in a common class?

web/src/main/webapp/heatmap3d.ts
```diff
@@ -0,0 +1,804 @@
...
+                result = "";
+        }
+        else if (kind == "Integer" || kind == "Double")
+            result = significantDigits(<number> inv);
+        // For Date do nothing
+        return result;
+    }
+}
+
+export class AxisData {
```
mbudiu-vmw: Can you merge this and HeatMap.ts AxisData into one class?

web/src/main/webapp/heatmap3d.ts
```diff
@@ -0,0 +1,804 @@
...
+            scale.range([length, 0]);
+            return [d3.axisLeft(scale), resultScale];
+        }
+    }
+}
+
+export class HeatMap3DView extends RemoteObjectView implements IScrollTarget {
+    protected topLevel: HTMLElement;
+    public static readonly maxColumns = 4;
+    public static readonly chartWidth = 200;  // pixels
```
mbudiu-vmw: Please reuse Resolution.

web/src/main/webapp/heatmap3d.ts
```diff
@@ -0,0 +1,804 @@
...
+    combine(how: CombineOperators): void {
+        let r = SelectedObject.current.getSelected();
+        if (r == null) {
+            this.page.reportError("No view selected");
+            return;
+        }
+
+        let rr = this.createRpcRequest("zip", r.remoteObjectId);
+        let renderer = (page: FullPage, operation: ICancellable) => {
+            return new Make2DHistogram(
```
mbudiu-vmw: Is this correct or do you want a Make3DHistogram class?

web/src/main/webapp/heatmap3d.ts
```diff
@@ -0,0 +1,804 @@
...
+                this.tableSchema, true);
+        };
+        rr.invoke(new ZipReceiver(this.getPage(), rr, how, renderer));
+    }
+
+    public static logScale(max: number) {
+        return max > 20;
+    }
+
+    public static scale(max: number, logScale: boolean) {
```
mbudiu-vmw: This stuff is also in common with HeatMapView, maybe you can refactor it into a common base class
HeatMapViewBase

web/src/main/webapp/heatmap3d.ts
```diff
@@ -0,0 +1,804 @@
...
+        page.setDataView(table);
+        this.page.insertAfterMe(page);
+        rr.invoke(new TableRenderer(page, table, rr, false, order));
+    }
+
+    getHTMLRepresentation(): HTMLElement {
+        return this.topLevel;
+    }
+
+    // Generates a string that encodes a call to the SVG translate method
```
mbudiu-vmw: Use the function from ui.ts

web/src/main/webapp/heatmap3d.ts
```diff
@@ -0,0 +1,804 @@
...
+            isAscending: true
+        }]);
+        let rr = table.createNextKRequest(order, null);
+        let page = new FullPage();
+        page.setDataView(table);
+        this.page.insertAfterMe(page);
+        rr.invoke(new TableRenderer(page, table, rr, false, order));
+    }
+
+    getHTMLRepresentation(): HTMLElement {
```
mbudiu-vmw: this is already inherited from RemoteObjectView, you don't need this method.

web/src/main/webapp/heatmap3d.ts
```diff
@@ -0,0 +1,804 @@
...
+            this.currentData.data,
+            this.currentData.xData,
+            this.currentData.yData,
+            this.currentData.zData,
+            this.currentData.missingData,
+            0);
+    }
+
+    public pageDown() {
+        console.log("TODO: Implement pageDown");
```
mbudiu-vmw: I am confused, I thought that I saw code for making pageDown work. Is this needed?

web/src/main/webapp/heatMapArray/types.ts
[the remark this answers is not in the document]
HanKruiger: I felt like the `webapp` folder was becoming a bit bloated, so I figured to make a folder for the code specific for the heatmap array. But I suppose a separation in terms of functionality (i.e., rpc stuff, ui stuff, etc.) is better than a separation in terms of user features (i.e., heatmap3d, pca, histograms, etc.)? Or do you want to just keep everything in the `webapp` folder?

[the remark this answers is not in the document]
mbudiu-vmw: This code will further be cleaned-up as it evolves. I have no strict preference on folders yet.

web/src/main/webapp/table.ts
HanKruiger: The problem here is that `this.selectedColumns` is a set, not an array. But I added functions for both cases in `util.ts` now, so it's simpler.

HanKruiger:
In the latest commits I've addressed most of your comments. Except for:

* Using the ColorMap in other views. I could do this in a different PR, if I find the time.
* The chart size that we aim for for the heat maps is still hard-coded. But I did make it responsive in the sense that the maximum number of heat maps visible is a function of the page width. Try it out if you want, it looks quite natural.

mbudiu-vmw:
I will merge this separately and push the merged version to master.
Then I will close this PR.

PR #168, opened 2017-08-28
Fix keyboard shortcuts in dialog

HanKruiger:
It didn't work earlier (for me at least). ~~The `onkeydown` has to be set in the global `document` element, since [otherwise the escape key isn't caught](https://stackoverflow.com/q/3369593/2501474)~~.


web/src/main/webapp/dialog.ts
```diff
@@ -68,11 +66,13 @@ export class Dialog implements IHtmlElement {
     // display the menu
     public show(): void {
         document.body.appendChild(this.container);
+        document.onkeydown = (ev) => this.handleKeypress(ev);
```
mbudiu-vmw: I don't really like how these handle a global document property when all we want is handling the key in the current page. This will probably only work correctly for modal menus, which have to disappear before anything else can be done.

PR #176, opened 2017-09-08
Map a column to a categorical column

HanKruiger:
For this to be possible, I also had to implement the CategoricalArrayColumn, which is also in the PR.


platform/src/main/java/org/hillview/maps/ToCatMap.java
```diff
@@ -0,0 +1,67 @@
...
+public class ToCatMap implements IMap<ITable, ITable> {
+    private final String inputColName;
+    private final String catColName;
+
+    /**
+     * @param inputColName The name of the column that has to be converted to a categorical column.
+     */
+    public ToCatMap(String inputColName) {
+        this.inputColName = inputColName;
+        this.catColName = inputColName + " (Cat.)";
```
mbudiu-vmw: The output column name should be generated by the UI and be a parameter.
Maybe you already have a column with this name. You have to make sure that the name is unique.

platform/src/main/java/org/hillview/maps/ToCatMap.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    public ITable apply(ITable table){
```
mbudiu-vmw: space before brace. The first empty line is not really necessary

platform/src/main/java/org/hillview/maps/ToCatMap.java
```diff
@@ -0,0 +1,67 @@
...
+        for (IColumn inputColumn : originalColumns) {
+            if (inputColumn.getName().equals(this.inputColName))
+                index = i;
+            columns.add(inputColumn);
+            i++;
+        }
+
+        // Get the column we want to map to a categorical column
+        IColumn inputColumn = table.getColumn(this.inputColName);
+        if (inputColumn.getDescription().kind == ContentsKind.Category)
```
mbudiu-vmw: Actually the column must be a String column for this to make sense.

platform/src/main/java/org/hillview/table/CategoryArrayColumn.java
```diff
@@ -0,0 +1,65 @@
+package org.hillview.table;
+
+import org.hillview.table.api.*;
+
+import javax.annotation.Nullable;
+import java.util.HashMap;
+
+public class CategoryArrayColumn extends BaseArrayColumn implements IStringColumn, IMutableColumn {
+
```
mbudiu-vmw: extra empty line

platform/src/main/java/org/hillview/utils/TestTables.java
```diff
@@ -54,7 +54,7 @@ public static Table testTable() {
      * @return A small table with some repeated content.
      */
     public static Table testRepTable() {
-        ColumnDescription c0 = new ColumnDescription("Name", ContentsKind.Category, false);
+        ColumnDescription c0 = new ColumnDescription("Name", ContentsKind.String, false);
```
mbudiu-vmw: Maybe you can avoid modifying this table, it is used in many places.
I guess no one depends on it being categorical.
Maybe you can change the column from StringArray to CategoryArray instead.

platform/src/main/java/org/hillview/maps/ToCatMap.java
HanKruiger: It could also make sense for integer columns. For example, the MNIST dataset might have had the labels (digits 0..9) in the table as integers, not categories. In general, every column should be convertible to a string (right?) so that makes it also convertible to a category (as long as the number of unique items is small enough to make sense).

mbudiu-vmw: Yes, we eventually will have almost all-to-all conversions.
But for numbers this will be done differently - probably by calling Integer.toString().

platform/src/main/java/org/hillview/utils/TestTables.java
HanKruiger: Sure, I'll make a new test table with a String array.
> Maybe you can change the column from StringArray to CategoryArray instead.
I don't follow. The column (in the testRepTable) used to be a String before I changed it in my branch. Do you want me to change it to a categorical column? That's what I did!?

mbudiu-vmw: The kind was Category but the implementation was StringColumn.
Use your new column instead, and don't change the kind.

HanKruiger: Ok, so it stays like this:
```
ColumnDescription c0 = new ColumnDescription("Name", ContentsKind.String, false);
```
That way the implementation stays StringColumn, as it was before the new Category implementation. I'll also change the `testTable()` (without the `Rep`) column in the same class.

(I see I switched the kinds around in my previous comment.)

mbudiu-vmw: No keep c0 unchanged, but make sac be instead a CategoricalColumn

HanKruiger: Oh, right. Now I understand. Sorry for the confusion and my slow morning brain. ;-)

HanKruiger:
I've processed your comments. The new map is a generic conversion of any column now. But we just support conversions from `String` to `Category` for now.

I'll do the front-end in a separate PR.

HanKruiger:
I've pushed the changes we made, with the frontend of column conversions working now.

PR #183, opened 2017-09-13
(Categorical) centroids sketch

HanKruiger:


HanKruiger:
I've changed it so that the final centroids are computed after the sketch finishes. So this can be reviewed/merged now.

platform/src/main/java/org/hillview/sketches/Centroids.java
```diff
@@ -0,0 +1,101 @@
...
+ * large, as they are shipped over the network.
+ */
+public class Centroids implements IJson {
+    /**
+     * Encodes the position of the centroids. Every row is a centroid, and the columns are the dimensions.
+     * Each partition has a centroid, which is simply the mean of the nD points in the partition.
+     * Should be computed with computeCentroids() by the head node.
+     */
+    @Nullable
+    public DoubleMatrix centroids;
```
mbudiu-vmw: Does this serialize as JSON properly?

platform/src/main/java/org/hillview/sketches/Centroids.java
```diff
@@ -0,0 +1,101 @@
...
+import org.jblas.ranges.PointRange;
+
+import javax.annotation.Nullable;
+import java.util.List;
+
+/**
+ * This class represents a set of centroids of a partitioning of a table. The number of centroids should not be too
+ * large, as they are shipped over the network.
+ */
+public class Centroids implements IJson {
```
mbudiu-vmw: You have here a bit too much; when you send the data to the client you don't actually want to serialize everything, only the actual centroids. Maybe you should not implement IJson at all, you just return from the web server the result of the call of computeCentroids.

platform/src/main/java/org/hillview/sketches/Centroids.java
```diff
@@ -0,0 +1,101 @@
...
+     * Each partition has a centroid, which is simply the mean of the nD points in the partition.
+     * Should be computed with computeCentroids() by the head node.
+     */
+    @Nullable
+    public DoubleMatrix centroids;
+    /**
+     * Holds the sum of the values in the partitions. Every row is the sum of all rows belonging to the corresponding
+     * partition.
+     */
+    public DoubleMatrix sums;
```
mbudiu-vmw: This requires you to have a numbering of the clusters into dense integers (the rows of this matrix). I think that it would be better to have a cluster be described by a key of some type T, and the sums should be a HashMap<T, DoubleVector>. Then the centroids of categories can use the category string for the keys, so it will be a Centroids<String> subclass.

platform/src/main/java/org/hillview/sketches/Centroids.java
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Registers how many rows are represented by every centroid.
+     */
+    public long[] counts;
```
mbudiu-vmw: Similarly, this will become a HashMap<T, long>

platform/src/main/java/org/hillview/sketches/Centroids.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Construct a centroids object that has processed the sums of the partitions in 'table'.
+     * The final centroids are not computed yet, this should be done after the sketch finishes.
+     * @param table The ITable that holds the data from which to compute the centroids.
+     * @param partitions The list of IMembershipSets that define the partitions.
+     * @param columnNames The list of column names of dimensions that define the nD space.
+     */
+    public Centroids(ITable table, List<IMembershipSet> partitions, List<String> columnNames) {
```
mbudiu-vmw: Here instead of the partitions list you take a function that computes the key given the row number: Function<Integer, T>

platform/src/main/java/org/hillview/sketches/Centroids.java
```diff
@@ -0,0 +1,101 @@
...
+        this.counts = new long[partitions.size()];
+        this.sums = DoubleMatrix.zeros(partitions.size(), columnNames.size());
+
+        int i = 0;
+        for (IMembershipSet partition : partitions) {
+            // The count is simply the number of rows that the sum represents.
+            this.counts[i] = partition.getSize();
+            if (this.counts[i] > 0) {
+                // Consider only part of the data that is in the partition.
+                ITable partitionTable = table.selectRowsFromFullTable(partition);
```
mbudiu-vmw: There is a lot of copying going on here.
But if you change to HashMaps this will go away.
Unlike the case of PCA, I don't think there is much benefit of using Blas here. It's just a linear-time operation.

platform/src/main/java/org/hillview/sketches/Centroids.java
```diff
@@ -0,0 +1,101 @@
...
+        return this;
+    }
+
+    /**
+     * Computes the centroids, based on the information in the 'sums' and 'counts' fields.
+     * This only has to be done once: when the sketch has finished.
+     * Sets the centroids in the objects 'centroids' field.
+     * @return The computed centroids.
+     */
+    public DoubleMatrix computeCentroids() {
```
mbudiu-vmw: This would also return a HashMap<T, DoubleVector>.
Maybe you can make that in a different class and call this class HighDimensionalAverage or something similar.

platform/src/main/java/org/hillview/sketches/Centroids.java
```diff
@@ -0,0 +1,101 @@
...
+
+        int i = 0;
+        for (IMembershipSet partition : partitions) {
+            // The count is simply the number of rows that the sum represents.
+            this.counts[i] = partition.getSize();
+            if (this.counts[i] > 0) {
+                // Consider only part of the data that is in the partition.
+                ITable partitionTable = table.selectRowsFromFullTable(partition);
+                // Get the matrix, compute the sum of every column, and put it in the sums matrix.
+                DoubleMatrix tableData = BlasConversions.toDoubleMatrix(partitionTable, columnNames);
```
mbudiu-vmw: what happens if you have missing values?
You have to document this in the comment above this class.

platform/src/main/java/org/hillview/sketches/Centroids.java
```diff
@@ -0,0 +1,101 @@
...
+     * Registers how many rows are represented by every centroid.
+     */
+    public long[] counts;
+
+    /**
+     * Constructs a zero-centroids. Needs to know the dimensions.
+     * @param numPartitions Number of partitions.
+     * @param numColumns Number of columns (= number of dimensions in the centroid space).
+     */
+    public Centroids(int numPartitions, int numColumns) {
```
mbudiu-vmw: Another benefit is that you don't need to know the number of partitions anymore; any key that does not exist in the hashmap implies a zero doublevector.

platform/src/main/java/org/hillview/sketches/Centroids.java
HanKruiger: DoubleMatrix implements Serializable, so it should, right? Anyway, it doesn't matter now, since I've refactored it to not use the JBLAS classes. Will push the changes when I finish updating the doc.

platform/src/main/java/org/hillview/sketches/Centroids.java
HanKruiger: It would make sense to make this a static method of `Centroid`, right? Because of how coupled it is with `Centroid`'s implementation.

platform/src/main/java/org/hillview/sketches/Centroids.java
HanKruiger: Oh right, so if I understand correctly, objects that have to be passed between backend nodes should implement just `Serializable`, while objects that have to be passed to the client (i.e., browser) should implement `IJson` (which also implements `Serializable`).

platform/src/main/java/org/hillview/sketches/Centroids.java
HanKruiger: I see now that this is not immediately evident. We should refrain from sending `DoubleMatrix` objects to the client anyway, so this is fine.

HanKruiger:
I've pushed another commit in which I've processed your requests!

* `CentroidsSketch` doesn't have to be abstract anymore, and `CategoricalCentroidSketch` has become quite small/trivial, which should be a good thing, right? That means we've generalized nicely, at least for this case.
* No more `DoubleMatrix` fields in `Centroids`.
* Less redundant copying.

platform/src/main/java/org/hillview/sketches/Centroids.java
mbudiu-vmw: IJson does not extend Serializable; Serializable means "Serialize the object the way Java knows", while IJson means "Serialize the object as JSON".
Unfortunately the IJson interface is not first-class like Serializable is, so just inheriting from IJson does not make it work automatically; sometimes you may need to do additional work, like registering some serialization/deserialization classes in the IJson file (take a look).

platform/src/main/java/org/hillview/sketches/Centroids.java
```diff
@@ -0,0 +1,105 @@
...
+     */
+    public Centroids() {
+        this.sums = new HashMap<T, double[]>();
+        this.counts = new HashMap<T, long[]>();
+    }
+
+    /**
+     * Construct a centroids object that has processed the sums of the partitions in 'table'.
+     * The final centroids are not computed yet, this should be done after the sketch finishes.
+     * @param members MembershipSet that has knows which rows to iterate over.
```
mbudiu-vmw: typo in comment "has knows"

platform/src/main/java/org/hillview/sketches/Centroids.java
```diff
@@ -0,0 +1,105 @@
...
+        for (IColumn column : columns) {
+            IRowIterator rowIterator = members.getIterator();
+            int row = rowIterator.getNextRow();
+            while (row >= 0) {
+                if (column.isMissing(row))
+                    continue;
+                T key = keyFunc.apply(row);
+                // Update the values. (And first create them if absent.)
+                this.sums.computeIfAbsent(key, k -> new double[columns.size()])[colIndex] += column.getDouble(row);
+                this.counts.computeIfAbsent(key, k -> new long[columns.size()])[colIndex]++;
```
mbudiu-vmw: this is a bit too compact, I had trouble parsing this.

platform/src/main/java/org/hillview/sketches/Centroids.java
```diff
@@ -0,0 +1,105 @@
...
+                    // this.count must also contain the key as well, as it shares key set with this.sum.
+                    this.counts.get(key)[colIndex] += count[colIndex];
+                }
+            } else {
+                // No need to iterate/sum. Simply use the other's array.
+                this.sums.put(key, other.sums.get(key));
+                this.counts.put(key, other.counts.get(key));
+            }
+        });
+        return this;
```
mbudiu-vmw: Please don't mutate this; we want everything to be immutable and thread-safe; allocate a new Centroids.
Also, it looks like you are losing elements in this which are not in other.

platform/src/main/java/org/hillview/sketches/Centroids.java
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * Constructs a zero-centroids.
+     */
+    public Centroids() {
+        this.sums = new HashMap<T, double[]>();
+        this.counts = new HashMap<T, long[]>();
+    }
+
+    /**
+     * Construct a centroids object that has processed the sums of the partitions in 'table'.
```
mbudiu-vmw: Please describe what happens for missing values.
Your treatment of missing values seems to be good.

HanKruiger:
Alright, I've processed the final remarks, so I think this can be merged now!

PR #188, opened 2017-09-19
Added ability to change color maps

HanKruiger:
The `ColorMap` and `ColorLegend` classes are now used in `HeatMap` and `HeatMapArray`. It would be nice to also use these in `Histogram2D`, but this is more challenging, as it should then also support selection in the legend.

### DIFF
--- a/web/src/main/webapp/heatMap.ts
+++ b/web/src/main/webapp/heatMap.ts
@@ -32,6 +32,7 @@ import {BaseType} from "d3-selection";
 import {ScaleLinear, ScaleTime} from "d3-scale";
 import {TopMenu, TopSubMenu} from "./menu";
 import {Histogram2DRenderer, Make2DHistogram, Filter2DReceiver} from "./histogram2d";
+import {ColorMap, ColorLegend} from "./vis";
 
 // counterpart of Java class 'HeatMap'
 export class HeatMapData {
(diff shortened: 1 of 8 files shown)


mbudiu-vmw:
We'd like to support selection in the legend for these too!

web/src/main/webapp/heatMap.ts
```diff
@@ -391,10 +326,28 @@ export class HeatMapView extends RemoteTableObjectView {
...
+            .attr("data-val", d => d.v)
             .attr("width", this.pointWidth)
             .attr("height", this.pointHeight)
             .style("stroke-width", 0)
-            .style("fill", d => this.color(d.v, max));
+            .style("fill", d => this.colorMap.apply(d.v));
+
+        this.colorLegend.redraw();
+        this.colorLegend.setColorMapChangeEventListener((cm) => {
+            this.chart.selectAll(".heatMapCell")
```
mbudiu-vmw: Can you turn this code into a method of 'this'?

web/src/main/webapp/menu.ts
```diff
@@ -30,14 +30,16 @@ export class ContextMenu implements IHtmlElement {
         this.outer = document.createElement("table");
         this.outer.classList.add("dropdown");
         this.outer.classList.add("menu");
-        this.outer.onmouseleave = () => {this.clear(); this.hide()};
+        // this.outer.onmouseleave = () => {this.clear(); this.hide()};
```
mbudiu-vmw: Please delete the unused code.

HanKruiger:
I've pushed the changes you requested!

Selection in the color legend can be done similarly as how the new colormap selection is done. With an `addSelectionListener` method that triggers the listener when a selection is made in the color legend. But this is future work.

PR #198, opened 2017-09-27
Hide submenus when leaving a top menu cell

HanKruiger:
That was a quicker fix than expected.

### DIFF
--- a/web/src/main/webapp/menu.ts
+++ b/web/src/main/webapp/menu.ts
@@ -88,7 +88,6 @@ export class TopSubMenu implements IHtmlElement {
         this.outer = document.createElement("table");
         this.outer.classList.add("menu", "hidden");
         this.tableBody = this.outer.createTBody();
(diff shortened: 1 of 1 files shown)


web/src/main/webapp/menu.ts
```diff
@@ -88,7 +88,6 @@ export class TopSubMenu implements IHtmlElement {
         this.outer = document.createElement("table");
         this.outer.classList.add("menu", "hidden");
         this.tableBody = this.outer.createTBody();
-        this.outer.onmouseleave = () => this.hide();
```
mbudiu-vmw: This line is not needed?

HanKruiger: No it isn't needed, it's handled in the TopMenu cell now. The cell in the TopMenu contains the submenu (in the DOM). When the mouse leaves that cell (or any of its children), the submenu disappears. This is actually the same behaviour as many other top menu's in applications, as far as I know.

PR #210, opened 2017-10-04
A few small changes for the demo

HanKruiger:
* Show the LAMP operation time on the page.
* Automatically enable log scale in LAMP view.
* Deterministic sampling
* Display on the page how many numeric columns were selected

### DIFF
--- a/web/src/main/webapp/lamp.ts
+++ b/web/src/main/webapp/lamp.ts
@@ -36,12 +36,15 @@ class ControlPointsView extends RemoteTableObjectView {
     private minY: number;
     private maxX: number;
     private maxY: number;
+    private maxVal: number; // Maximum value in the heat map
     private heatMapCanvas: any;
     private heatMapChart: any;
     private controlPointsCanvas: any;
     private controlPointsChart: any;
     public controlPoints: PointSet2D;
-    private heatMap: HeatMapData;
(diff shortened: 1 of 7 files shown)


web/src/main/webapp/lamp.ts
```diff
@@ -295,12 +305,13 @@ export class LAMPDialog extends Dialog {
...
                 private schema: Schema, private remoteObject: TableView) {
         super("LAMP");
-        this.addTextField("numSamples", "No. control points", "Integer", "15");
+        this.addTextField("numSamples", "No. control points", "Integer", "20");
         this.addSelectField("controlPointSelection", "Control point selection", ["Random samples", "Category centroids"], "Random samples");
         let catColumns = [""];
         for (let i = 0; i < schema.length; i++)
             if (schema[i].kind == "Category")
                 catColumns.push(schema[i].name);
+        this.addTextField("seed", "Random seed", "Integer", "1");
```
mbudiu-vmw: I am not sure you really want the user to give you this. 
You can generate it locally somehow. The point is that we can re-execute the message later and it will generate the same result. The message will contain the seed.

mbudiu-vmw: I will merge this, you can make the seed change later.

PR #220, opened 2017-10-10
TopK performance improvements

lalithsuresh:
Reduces allocations in the TopK implementations and NextKSketch. Given that all uses of `TreeTopK<T>` were with `TreeTopK<Integer>`, I added a specialized version `IntTreeTopK`. Allocations all in all were reduced from about 115MB -> ~600KB per TopK against 6.6M records.

### DIFF
--- a/platform/src/main/java/org/hillview/sketches/IntTreeTopK.java
+++ b/platform/src/main/java/org/hillview/sketches/IntTreeTopK.java
@@ -17,60 +17,70 @@
 
 package org.hillview.sketches;
 
+import it.unimi.dsi.fastutil.ints.Int2IntRBTreeMap;
+import it.unimi.dsi.fastutil.ints.Int2IntSortedMap;
+import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
+import it.unimi.dsi.fastutil.ints.Int2ObjectRBTreeMap;
+import it.unimi.dsi.fastutil.ints.IntComparator;
 import it.unimi.dsi.fastutil.objects.Object2IntRBTreeMap;
 import it.unimi.dsi.fastutil.objects.Object2IntSortedMap;
 import it.unimi.dsi.fastutil.objects.Object2ObjectRBTreeMap;
+import it.unimi.dsi.fastutil.objects.Object2ObjectSortedMap;
 import org.hillview.utils.MutableInteger;
 
 import javax.annotation.Nullable;
 import java.util.Comparator;
 import java.util.SortedMap;
+import java.util.TreeMap;
 
 /**
- * Implements the ITopK interface as a SortedMap (which uses Red-Black trees).
- * Seems faster than HashMap implementation.
+ * Implements an int specialized version of the ITopK interface. Reduces boxing
+ * overheads in comparison to the TreeTopK class.
  */
-public class TreeTopK<T> implements ITopK<T> {
+public class IntTreeTopK implements ITopK<Integer> {
     private final int maxSize;
     private int size;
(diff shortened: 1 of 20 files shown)


platform/src/main/java/org/hillview/sketches/IntTreeTopK.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public Int2IntSortedMap getTopK() {
+        final Int2IntSortedMap finalMap = new Int2IntRBTreeMap(this.greater);
+        this.data.forEach((k, v) -> finalMap.put(k.intValue(), v.get()));
+        return finalMap;
+    }
+
+    @Deprecated
+    @Override
+    public void push(final Integer newVal) {
```
mbudiu-vmw: why add a feature which is deprecated from the start?

platform/src/main/java/org/hillview/sketches/IntTreeTopK.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+public class IntTreeTopK implements ITopK<Integer> {
+    private final int maxSize;
+    private int size;
+    private final Int2ObjectRBTreeMap<MutableInteger> data;
+    private int cutoff; /* max value that currently belongs to Top K. */
+    private final IntComparator greater;
+
+    public IntTreeTopK(final int maxSize, final IntComparator greater) {
+        this.maxSize = maxSize;
```
mbudiu-vmw: Want to check for the maxSize being > 0?

platform/src/main/java/org/hillview/table/ListComparator.java
```diff
@@ -19,19 +19,20 @@
...
+    public ListComparator(final ArrayList<IndexComparator> comparatorList) {
         this.comparatorList = comparatorList;
     }
 
     @Override
-    public int compare(final Integer o1, final Integer o2) {
-        for (final IndexComparator aComparator : this.comparatorList) {
-            final int val = aComparator.compare(o1, o2);
+    public int compare(final int o1, final int o2) {
+        for (int i = 0; i < this.comparatorList.size() ; i++) {
```
mbudiu-vmw: That's funny, I would expect the other iterator to be more efficient.

platform/src/main/java/org/hillview/table/ListComparator.java
```diff
@@ -19,19 +19,20 @@
 
 import org.hillview.table.api.IndexComparator;
 
+import java.util.ArrayList;
 import java.util.List;
 
 public class ListComparator extends IndexComparator {
-    private final List<IndexComparator> comparatorList;
+    private final ArrayList<IndexComparator> comparatorList;
```
mbudiu-vmw: Why does this make a difference?

mbudiu-vmw:
I will merge this anyway.

platform/src/main/java/org/hillview/sketches/IntTreeTopK.java
lalithsuresh: The intent here was to implement the ITopK<> interface (not the different parameter type). Doing so satisfies the interface, but warns us via compiler warnings when the wrong push() is being used.

platform/src/main/java/org/hillview/table/ListComparator.java
lalithsuresh: Perhaps not the best way to do this, but I wanted it to be clear that we intend to directly index into the array when iterating through it. Otherwise, we create a lot of Iterator garbage.

PR #224, opened 2017-10-11
Cietro

uwieder:
Added a small tabulated file with a .gz extension to the data folder. Created a button for it and set the CsvFileReader to detect the gz extension and use the appropriate reader.


data/criteo.txt
```diff
@@ -0,0 +1,1000 @@
+1	5	110		16		1	0	14	7	1		306		62770d79	e21f5d58	afea442f	945c7fcf	38b02748	6fcd6dcb	3580aa21	28808903	46dedfa6	2e027dc1	0c7c4231	95981d1f	00c5ffb7	be4ee537	8a0b74cc	4cdc3efa	d20856aa	b8170bba	9512c20b	c38e2f28	14f65a5d	25b1b089	d7c1fc0b	7caf609c	30436bfc	ed10571d
```
mbudiu-vmw: Why do we have four different files with the same data?
Only one seems to be necessary.

platform/src/main/java/org/hillview/storage/CsvGzReader.java
```diff
@@ -0,0 +1,144 @@
+package org.hillview.storage;
```
mbudiu-vmw: It looks like we don't need this class at all.

web/src/main/java/org/hillview/InitialObjectTarget.java
```diff
@@ -136,9 +136,12 @@ void prepareFiles(RpcRequest request, RpcRequestContext context) {
         } else if (which == 4) {
             fileNamePattern = "segmentation.csv";
             schemaFile = "segmentation.schema";
-        } else {
+        } else if (which == 5) {
```
mbudiu-vmw: please use 4 spaces of indentation

web/src/main/webapp/npm-debug.log
```diff
@@ -0,0 +1,112 @@
+0 info it worked if it ends with ok
```
mbudiu-vmw: Please delete this file

platform/src/main/java/org/hillview/storage/CsvGzReader.java
uwieder: Correct, I thought I removed it from git, apparently I didn't.

web/src/main/java/org/hillview/InitialObjectTarget.java
```diff
@@ -136,10 +136,14 @@ void prepareFiles(RpcRequest request, RpcRequestContext context) {
...
             schemaFile = "segmentation.schema";
+        } else if (which == 5) {
+            config.separator = '\t';
+            fileNamePattern = "criteoTab.gz";
+            schemaFile = "criteo.schema";
         } else {
             throw new RuntimeException("Unexpected operation " + which);
-        }
-
+		}
```
mbudiu-vmw: Please don't use tabs. You should configure your editor to always use spaces.
If it is emacs I can send you some lisp code.

PR #225, opened 2017-10-13
Compressed category column

uwieder:
Implemented a switch from a byte encoding to int. I saw some savings in memory using top. Further improvement possible by switching first to short and then to int.


platform/src/main/java/org/hillview/table/columns/CategoryEncoding.java
```diff
@@ -28,30 +31,52 @@
  */
 public class CategoryEncoding {
     // Map categorical value to a small integer
-    private final Object2IntOpenHashMap<String> encoding;
+    private final Object2IntOpenHashMap<String> intEncoding;
     // Decode small integer into categorical value
-    private final Int2ObjectOpenHashMap<String> decoding;
+    private final Int2ObjectOpenHashMap<String> intDecoding;
+    // Map categorical value to a byte
+    private final Object2ByteOpenHashMap<String> byteEncoding;
```
mbudiu-vmw: I don't think you need to change this class. There will be no significant savings by doing this.

platform/src/main/java/org/hillview/table/columns/CategoryListColumn.java
```diff
@@ -31,42 +31,67 @@
  */
 public class CategoryListColumn extends BaseListColumn implements ICategoryColumn {
     private CategoryEncoding encoding;
-    private final ArrayList<int[]> segments;
+    private final ArrayList<int[]> intSegments;
+    private final ArrayList<byte[]> byteSegments;
+    private int switchPoint; //the row index beyond which int segments are used
```
mbudiu-vmw: This can probably be a segmentId.

PR #226, opened 2017-10-14
Benchmarking

lalithsuresh:
Adds a runnable jar that benchmarks histograms. Also involves removing several null checks.


platform/src/main/java/org/hillview/table/api/IStringColumn.java
```diff
@@ -18,19 +18,18 @@
...
 public interface IStringColumn extends IColumn {
     @Override
-    default double asDouble(final int rowIndex, @Nullable final IStringConverter converter) {
+    default double asDouble(final int rowIndex, final IStringConverter converter) {
         if (isMissing(rowIndex))
             throw new MissingException(this, rowIndex);
         final String tmp = this.getString(rowIndex);
         //noinspection ConstantConditions
-        return Converters.checkNull(converter).asDouble(tmp);
+        return converter.asDouble(tmp);
```
mbudiu-vmw: the comment above it probably unnecessary

platform/src/main/java/org/hillview/table/columns/BaseArrayColumn.java
```diff
@@ -29,24 +27,24 @@
...
             throw new InvalidParameterException("Size must be positive: " + size);
-        if (this.description.allowMissing && !this.description.kind.isObject())
-            this.missing = new BitSet(size);
+        this.missing = new BitSet(size);
     }
 
     @Override
     public boolean isMissing(final int rowIndex) {
-        return this.description.allowMissing && Converters.checkNull(this.missing).get(rowIndex);
+        // Ignore the analysis complaint. this.missing can not be null.
```
mbudiu-vmw: Why is it complaining?
There is an official IDEA comment for this.
Same observation below

platform/src/main/java/org/hillview/table/NoOpStringConverter.java
```diff
@@ -0,0 +1,31 @@
...
+import org.hillview.table.api.IStringConverter;
+import org.hillview.table.api.IStringConverterDescription;
+
+import javax.annotation.Nullable;
+
+/**
+ * A string converter that does nothing
+ */
+public class NoOpStringConverter implements IStringConverter, IStringConverterDescription {
+    private static final NoOpStringConverter INSTANCE = new NoOpStringConverter();
```
mbudiu-vmw: Why not make this public and remove getInstance()?

platform/src/main/java/org/hillview/table/NoOpStringConverter.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.hillview.table.api.IStringConverter;
+import org.hillview.table.api.IStringConverterDescription;
+
+import javax.annotation.Nullable;
+
+/**
+ * A string converter that does nothing
+ */
+public class NoOpStringConverter implements IStringConverter, IStringConverterDescription {
```
mbudiu-vmw: let's call this NoStringConverter.
It's not a no-op, it's a throw.

platform/src/main/java/org/hillview/table/columns/BaseArrayColumn.java
lalithsuresh: Stale comment. Will remove.

platform/src/main/java/org/hillview/table/NoOpStringConverter.java
lalithsuresh: While not strictly necessary here, I did it out of habit (it's cleaner to access such singletons behind a method in the event we want to say, log every access to the API).

PR #227, opened 2017-10-14
More inner-loop optimizations

lalithsuresh:
This patch further optimizes our inner-most loops by removing redundant invocations of isMissing(). They have been replaced with asserts to catch bugs during development time and unit testing. 

The reason this works is because all our sketches check for isMissing() explicitly before invoking methods such asDouble().


platform/src/main/java/org/hillview/table/api/IDateColumn.java
```diff
@@ -27,17 +27,15 @@
 public interface IDateColumn extends IColumn {
     @Override
     default double asDouble(final int rowIndex, final IStringConverter unused) {
-        if (isMissing(rowIndex))
-            throw new MissingException(this, rowIndex);
+        assert !isMissing(rowIndex);
```
mbudiu-vmw: Maybe you can delete MissingException?

platform/src/main/java/org/hillview/table/SortedStringsConverterDescription.java
```diff
@@ -61,17 +61,22 @@ public IStringConverter getConverter() {
...
             this.memoizedResults = new Object2DoubleOpenHashMap<String>();
+            this.keyNotFound = (boundaries.length) + 2;
         }
 
         @Override
         public double asDouble(@Nullable String string) {
-            if (memoizedResults.containsKey(string)) {
-                return memoizedResults.getDouble(string);
+            double index = memoizedResults.getOrDefault(string, keyNotFound);
+            if (index < keyNotFound - EPSILON) {
```
mbudiu-vmw: I don't get the epsilon.

lalithsuresh: Was worried about floating point precision problems.

platform/src/main/java/org/hillview/table/api/IDateColumn.java
lalithsuresh: Will do.

platform/src/main/java/org/hillview/table/SortedStringsConverterDescription.java
mbudiu-vmw: there is no arithmetic on these fields, so that should not be a problem
we often compare doubles for equality, e.g. in heavy hitters or hyperLogLog

PR #229, opened 2017-10-16
Post-fixes from loop-opt

lalithsuresh:
Some tweaks to how we create executors. Also sets grpc/netty logging to INFO.


deployment/prepare.yaml
```diff
@@ -69,7 +69,6 @@
         creates: "{{service_folder}}/{{tomcat}}"
         src: "http://archive.apache.org/dist/tomcat/tomcat-8/v{{tomcat_version}}/bin/{{tomcat}}.tar.gz"
         dest: "{{service_folder}}"
-        copy: no
```
mbudiu-vmw: why are you deleting this?
Doesn't this mean that the download will happen every time?

lalithsuresh: No it doesn't. On my verison of ansible, it says that "copy" and "remote_src" are mutually exclusive, and the script fails unless that is removed.

PR #230, opened 2017-10-18
Fix latency issues associated with Epoll.

lalithsuresh:


platform/src/main/java/org/hillview/dataset/remoting/HillviewServer.java
```diff
@@ -72,7 +73,10 @@
     private static final int NUM_THREADS = 5;
     public static final int MAX_MESSAGE_SIZE = 20971520;
     private final ExecutorService executorService = ExecutorUtils.newNamedThreadPool("server", NUM_THREADS);
-    private final EventLoopGroup nettyElg = new NioEventLoopGroup(1);
+    private final EventLoopGroup workerElg = new NioEventLoopGroup(1,
```
mbudiu-vmw: how about some comments about what this stuff does and why?
with a link to a relevant web page?

PR #241, opened 2017-10-30
Iterate sample

uwieder:
Iterator over Sampled  membership maps for Full Membership and Sparse Membership


platform/src/main/java/org/hillview/utils/IntSet.java
```diff
@@ -97,6 +97,19 @@ public boolean add(final int k) {
         return true;
     }
 
+    /**
+     *
+     * @param pos a location in the key[] array
+     * @return in res[0] the location of the next full slot after cursor. In res[1] the value.
+     */
+    public int[] getNext(int pos) {
+        int[] res = new int[2];
```
mbudiu-vmw: This is expensive: you allocate an array for every call.
It's better to pass the array as an out argument, or to expose the internal indexing operation and return just the internal index as an opaque value.

platform/src/main/java/org/hillview/utils/Randomness.java
```diff
@@ -43,6 +43,21 @@ public Randomness(long seed) {
...
 
     /**
+     * @return the next pseudarandum Geometric with distribution parameter p. Should be more efficient than using logs.
+     */
+    public int nextGeometric(double p) {
+        if (p <= 0) throw new IllegalArgumentException("parameter of Geometric Distribution must be positive");
+        double sample = myPrg.nextDouble();
+        int i = 0;
+        double bar = 0;
+        while (sample > bar) {
```
mbudiu-vmw: This is a non-constant cost too.
What is the amortized cost to do this for a full FullMembership?

platform/src/main/java/org/hillview/utils/IntSet.java
uwieder: You are right, I made the change

platform/src/main/java/org/hillview/utils/Randomness.java
uwieder: No excuse for this brain fart. I rewrote it the right way.

PR #271, opened 2017-12-06
Sorted iterator

uwieder:
an optimized iterator for sparse list, should improve performance considerably. fixed a minor bug in the iterator and added performance benchmarks


platform/src/main/java/org/hillview/FactoryBenchmark.java
```diff
@@ -0,0 +1,70 @@
...
+import org.hillview.table.membership.SparseMembershipSet;
+import org.hillview.utils.HillviewLogger;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Level;
+
+public class FactoryBenchmark {
```
mbudiu-vmw: Please make this a test. The other benchmark files need to have a main for guaranteeing precision.
This one probably does not.

mbudiu-vmw: But if it is a test you may want to keep it disabled by default, because it takes a really long time.

platform/src/main/java/org/hillview/SamplingBenchmark.java
```diff
@@ -0,0 +1,104 @@
...
+                    buckDes, new ColumnAndConverterDescription(col.getName()), 1, 0);
+            Runnable r = () -> ds.blockingSketch(sk);
+            writer.print("1 , ");
+            runNTimes(r, runCount, "Dataset histogram wo sampling", elements, writer);
+            System.out.println(elements);
+        }
+        writer.close();
+    }
+
+    private static ITable createTable(final int colSize, final IColumn col) {
```
mbudiu-vmw: Can you share these methods with the previous class?
If they are tests it may be even easier.

platform/src/main/java/org/hillview/table/membership/SparseMembershipSet.java
```diff
@@ -32,6 +32,7 @@
  */
 public class SparseMembershipSet implements IMembershipSet, IMutableMembershipSet {
     private final IntSet membershipMap;
+    private static final int thresholdSortedIterator = 50000000; //this intSet is used for the sole purpose of iteration.
```
mbudiu-vmw: Please make this comment clearer.

platform/src/main/java/org/hillview/utils/IntSet.java
```diff
@@ -94,20 +98,21 @@ public boolean add(final int k) {
...
         }
+        this.iteratorKey = this.key;
         return true;
     }
 
     /**
      *
      * @param pos a location in the key[] array
-     * @return the location of the next full slot after cursor.
+     * @return the location of the next full slot after cursor. Operates on the iteratorKey array
```
mbudiu-vmw: In general you should not expose in such a comment information about a private field.
The comment for an API function should be read as describing what the function does, but not how it does it - unless it's really important for the caller to know this.

platform/src/main/java/org/hillview/utils/IntSet.java
```diff
@@ -94,20 +98,21 @@ public boolean add(final int k) {
...
      */
     public int getNext(int pos) {
-        while ( key[pos & this.mask] == 0) { pos++; }
+        while ( iteratorKey[pos & this.mask] == 0) { pos++; }
```
mbudiu-vmw: this.
Previously the code was copy-and-paste, but if you edit it you should also use our coding conventions.

platform/src/main/java/org/hillview/utils/IntSet.java
```diff
@@ -173,10 +179,20 @@ public IntSet copy() {
...
         System.arraycopy(this.key, 0, newSet.key, 0, this.key.length);
         return newSet;
     }
 
+    /**
+     * Creates another array with the keys, this time in sorted order. This is used to optimize the iterator and
+     * therefore should be called only once the IntSet is deemed immutable.
+     */
+    public void sortIterator() {
+        iteratorKey = Arrays.copyOf(this.key, this.key.length);
```
mbudiu-vmw: Then make this function private and call it from seal yourself.

platform/src/test/java/org/hillview/test/EqualityFilterTest.java
```diff
@@ -17,6 +17,7 @@
 
 package org.hillview.test;
 
+import org.hillview.dataset.LocalDataSet;
```
mbudiu-vmw: Why is this needed now?

platform/src/main/java/org/hillview/SamplingBenchmark.java
uwieder: No, there are slight differences in the implementaiton

platform/src/main/java/org/hillview/utils/IntSet.java
uwieder: It is important, because if the iterator is sorted this gives a sorted view which does not correspond to the hashing. This function exposes an internal array, so one may argue that in a perfect design this function is not implemented at all, but in our case it seems like a reasonable tradeoff for the optimization we do.

platform/src/test/java/org/hillview/test/EqualityFilterTest.java
uwieder: Just a leftover of the debugging process. Removed.

PR #277, opened 2017-12-19
Column selection

parikg:
Added user manual entry for schema browser


docs/userManual.md
```diff
@@ -287,18 +287,52 @@ marked with x.
...
 there are three rows selected.
 
 ![Schema view](schema-view.png)
 
-Once the user selects a set of columns, the user can display a view of the
-data table restricted to the selected columns using the View/Selected
-columns menu.
 
+There are two ways to modify the selection:
+1. By using the mouse: Clicking on a column description toggles its
```
mbudiu-vmw: The user manual already has a section describing how mouse-based selection works. I would rather point to that section here than write it again. It will also make it easier to change behaviors if we decide we don't like the current state-machine.

web/src/main/webapp/dataViews/equalityFilter.ts
```diff
@@ -15,48 +15,34 @@
...
-}
 
 export class EqualityFilterDialog extends Dialog {
     constructor(private columnDescription: ColumnDescription) {
         super("Filter on " + columnDescription.name, "Eliminates data from a column according to its value.");
         this.addTextField("query", "Find:", FieldKind.String, null, "Value to search");
-        this.addSelectField("complement", "Check for:", ["Equality", "Inequality"], null,
-            "For 'equality' the search will keep all values that are equal to the one chosen," +
-        "For 'inequality' the search will keep all values that are different from the one chosen.");
+        this.addBooleanField("asRegEx", "Interpret as Regular Expression", false, "Click " +
```
mbudiu-vmw: This should not say "click", but "if selected".

docs/userManual.md
```diff
@@ -469,9 +501,9 @@ the current state of the display.
 
 * Filter...: this option will pop-up a dialog window that allows the user
   to filter the data in the selected column (this option requires only
-  one column to be selected).  The user can specify a value to be
-  sought, and a comparison, indicating whether filtering keeps values
-  identical or different to the specified one.
+  one column to be selected).  The user can specify a value/pattern to be
+  searched for (regular expressions are permitted), and also whether the
```
mbudiu-vmw: I would not be so succinct. You can say that the pattern is interpreted as a regular expression using Java syntax if the next checkbox is selected.

PR #292, opened 2018-02-16
Cache

mbudiu-vmw:


mbudiu-vmw:
This seems to work pretty well. I tested it with a 20 seconds expiration policy.

web/src/main/java/org/hillview/RpcObjectManager.java
```diff
@@ -95,8 +95,13 @@ synchronized void addSubscription(RpcRequestContext context, Subscription subscr
             return;
         HillviewLogger.instance.info("Saving subscription", "{0}", context.toString());
         if (this.sessionSubscription.get(session) != null)
```
lalithsuresh: Nit: Use braces for consistency.

web/src/main/java/org/hillview/RpcObjectManager.java
```diff
@@ -162,10 +167,10 @@ private void rebuild(RpcTarget.Id id, Observer<RpcTarget> toNotify) {
         HillviewComputation computation = this.generator.get(id);
         if (computation != null) {
             // The following may trigger a recursive reconstruction.
-            HillviewLogger.instance.info("Replaying", "{0}", computation);
+            HillviewLogger.instance.info("Replaying", "computation={0}", computation);
             computation.replay(toNotify);
         } else {
-            Exception ex = new RuntimeException("Cannot reconstruct " + id);
+            Exception ex = new RuntimeException("Don't know how to reconstruct " + id);
```
lalithsuresh: What execution might lead to the 'else' branch?

platform/src/main/java/org/hillview/dataset/remoting/HillviewServer.java
```diff
@@ -102,26 +106,62 @@ public HillviewServer(final HostAndPort listenAddress, final IDataSet dataSet) t
...
+                        (RemovalListener<Integer, IDataSet>) removalNotification ->
+                                HillviewLogger.instance.info("Removing reference to dataset", "{0}: {1}",
+                                    removalNotification.getKey(), removalNotification.getValue().toString()))
+                .build();
     }
 
-    synchronized private IDataSet get(int index) {
-        return this.dataSets.get(index);
-    }
+    //////////////////////
```
lalithsuresh: Remove

platform/src/main/java/org/hillview/dataset/remoting/HillviewServer.java
```diff
@@ -102,26 +106,62 @@ public HillviewServer(final HostAndPort listenAddress, final IDataSet dataSet) t
...
+     * @return The number of deleted datasets.
+     */
+    synchronized public int deleteAllDatasets() {
+        long removed = this.dataSets.size();
+        this.dataSets.invalidateAll();
+        this.memoizedCommands.clear();
+        return (int) removed;
+    }
+
+    ////////////////////////
```
lalithsuresh: Remove

platform/src/main/java/org/hillview/dataset/remoting/HillviewServer.java
```diff
@@ -102,26 +106,62 @@ public HillviewServer(final HostAndPort listenAddress, final IDataSet dataSet) t
...
-    synchronized private void remove(int index) {
-        HillviewLogger.instance.info("Removing dataset", "{0}", index);
-        this.dataSets.remove(index);
+    /**
+     * Retrieve the dataset with the specified index.
+     * @param index     Dataset index.
+     * @param observer  Observer that is notified if the dataset is not available.
+     * @return          The dataset, or null if there is no such dataset.
+     */
+    @SuppressWarnings("BooleanMethodIsAlwaysInverted")
```
lalithsuresh: Looks like this is not required because there is no use of inverted booleans in the method

platform/src/main/java/org/hillview/dataset/remoting/HillviewServer.java
```diff
@@ -81,18 +79,24 @@
...
             = new ConcurrentHashMap<UUID, Subscription>();
     private final HostAndPort listenAddress;
-    private final ConcurrentHashMap<ByteString, Map<Integer, PartialResponse>> memoizedCommands
-            = new ConcurrentHashMap<ByteString, Map<Integer, PartialResponse>>();
+
+    private MemoizedResults memoizedCommands;
+
     @SuppressWarnings("CanBeFinal")
     private boolean MEMOIZE = true;
+    private static final int expireTimeInHours = 2;
```
lalithsuresh: make this EXPIRE_TIME_IN_HOURS.

web/src/main/java/org/hillview/RpcObjectManager.java
mbudiu-vmw: This should not happen, but I've seen many things that should not happen actually happen.

mbudiu-vmw: Actually, maybe there is a race when many threads remove the same object?
E.g., there is UI to remove objects...

mbudiu-vmw: Sorry, wrong response to your question. This can happen if a malicious browser client sends a request to access an object with an unknown guid. It can also happen currently because we rebooted the root server and we currently we do not save the redo log on persistent storage.

PR #293, opened 2018-02-17
Execute frontend RPC responses on a separate threadpool

lalithsuresh:
This hopefully addresses the race condition around interrupts.

### DIFF
--- a/web/src/main/java/org/hillview/RpcTarget.java
+++ b/web/src/main/java/org/hillview/RpcTarget.java
@@ -32,6 +32,8 @@ import javax.websocket.Session;
 import java.lang.reflect.Method;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


lalithsuresh:
I don't think this patch affects cancellation (unless we previously had a way to cancel the action of writing messages into a websocket while it is underway). Just that the replying is now done on a separate thread.

web/src/main/java/org/hillview/RpcTarget.java
```diff
@@ -98,6 +100,11 @@ public Id getId() {
         return this.objectId;
     }
 
+    /**
```
mbudiu-vmw: Maybe you can add a longer comment explaining why this is necessary.
No one will guess.

mbudiu-vmw:
You haven't really tested this, have you?
Everything stopped working - now the web socket is closed before sending the reply back...

mbudiu-vmw:
I will remove the replyExecutor for now. It does not work.

PR #309, opened 2018-03-15
Lazy loading for MySQL

mbudiu-vmw:


mbudiu-vmw:
More storage mechanisms will follow MySQL.
In fact, this may already work for Impala.
The next step is Parquet.
Also, more callers should use the getLoadedColumns API to ensure batch loading of columns.

PR #313, opened 2018-03-23
Manual update

uwieder:
updated from histogram 2d onwards


docs/userManual.md
```diff
@@ -888,7 +881,7 @@ axes of a plot, then the screen is divided into small patches a few
 pixels wide and the number of data points that falls within each patch
 is counted.  The number of values that falls within each patch is
 displayed as a heatmap, where the color intensity indicates the number
-of points.  A heatmap where the Y axis is not categorical will also
+of points. The mapping between the color intencity and the count could be in log scale if the range is large or linear if the range is smaller. Next to the mouse an overlay box displays the x value, the y value and the count. A heatmap where the Y axis is not categorical will also
```
mbudiu-vmw: the overlay box is already described earlier in this section.

platform/src/main/java/org/hillview/storage/JsonFileLoader.java
```diff
@@ -27,6 +27,7 @@
 import org.hillview.table.Schema;
 import org.hillview.table.Table;
 import org.hillview.table.api.*;
+import org.hillview.utils.HillviewLogger;
```
mbudiu-vmw: Why add this line?

docs/userManual.md
mbudiu-vmw: Can you spell-check this?

mbudiu-vmw: Can you check whether this text is duplicated in this section?

platform/src/main/java/org/hillview/storage/JsonFileLoader.java
uwieder: a mistake, I'll remove it.

docs/userManual.md
uwieder: Next paragraph it says that a menu can toggle between log scale in linear scale. Here we say that the default depends on the data range. I don't find any other duplication.

mbudiu-vmw: I am talking about the overlay box next to the mouse.

PR #367, opened 2018-07-02
Smart buckets

uwieder:
Will choose a number of buckets that would minimize bucketing effects. Note that this is true for all histograms. So in two dimensional histograms if one access has (say) a range of 44 and 40 is the maximum number of colors allowed, then it would choose 22 colors, each with two values.

### DIFF
--- a/web/src/main/webapp/dataViews/histogramViewBase.ts
+++ b/web/src/main/webapp/dataViews/histogramViewBase.ts
@@ -217,7 +217,8 @@ export abstract class HistogramViewBase extends BigTableView {
 
         if (columnKind == "Integer" ||
             columnKind == "Category")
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
Have you tried this for a variety of datasets? Does it behave as expected or is it sometimes surprising?
You can do this by selecting a dataset and performing various selections in a histogram.

uwieder:
Yes, I played around a bit and it seems to produce a more sensible outcome. The only downside is that the avoidance of combing can come at the expense of the number of buckets so we may want a smarter solution. For instance, if the number of item is 42 it would choose 21 buckets of load 2 each, deciding it is better than 40 buckets, 38 of which with one item and two with two items.

PR #375, opened 2018-07-16
String histogram

parikg:
Includes min and max computation in computing the sample of distinct elements.

### DIFF
--- a/platform/src/main/java/org/hillview/sketches/SampleDistinctElementsSketch.java
+++ b/platform/src/main/java/org/hillview/sketches/SampleDistinctElementsSketch.java
@@ -35,23 +35,35 @@ public class SampleDistinctElementsSketch implements ISketch<ITable, MinKSet<Str
             throw new IllegalArgumentException(
                     "SampleDistinctElementsSketch only supports String columns");
         LongHashFunction hash = LongHashFunction.xx(this.seed);
+        String minString = null;
+        String maxString = null;
         final IRowIterator myIter = data.getMembershipSet().getIterator();
         int currRow = myIter.getNextRow();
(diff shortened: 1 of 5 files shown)


platform/src/main/java/org/hillview/sketches/SampleDistinctElementsSketch.java
```diff
@@ -35,23 +35,35 @@ public MinKSet create(ITable data) {
...
         int currRow = myIter.getNextRow();
+        if (currRow >= 0) {
+            minString = col.getString(currRow);
+            maxString = col.getString(currRow);
+        }
         MinKRows mkRows = new MinKRows(this.maxSize);
         while (currRow >= 0) {
             if (!col.isMissing(currRow)) {
                 mkRows.push(col.hashCode64(currRow, hash), currRow);
+                if (minString.compareTo(col.getString(currRow)) > 0)
```
mbudiu-vmw: don't call getString twice

platform/src/main/java/org/hillview/sketches/SampleDistinctElementsSketch.java
```diff
@@ -67,12 +79,25 @@ public MinKSet create(ITable data) {
     @Nullable
     public MinKSet<String> add(@Nullable MinKSet<String> left, @Nullable MinKSet<String> right) {
         Long2ObjectRBTreeMap<String> data = new Long2ObjectRBTreeMap<>();
+        String minString, maxString;
+        if (left.min == null) {
+            minString = right.min;
+            maxString = right.max;
+        }
```
mbudiu-vmw: else should be on the same line as with the brace

PR #390, opened 2018-08-29
Substring filter

parikg:


platform/src/main/java/org/hillview/table/filters/StringFilter.java
```diff
@@ -0,0 +1,51 @@
...
+public class StringFilter implements IStringFilter {
+    private final StringFilterDescription stringFilterDescription;
+    boolean missing;  // if true we look for missing values;
+    @Nullable
+    String compareTo;
+    @Nullable
+    Pattern regEx;
+
+    public StringFilter(StringFilterDescription stringFilterDescription) {
+        org.junit.Assert.assertNotNull(stringFilterDescription);
```
mbudiu-vmw: use assert, not junit.Assert

platform/src/main/java/org/hillview/table/filters/StringFilter.java
```diff
@@ -0,0 +1,51 @@
...
+                this.compareTo = stringFilterDescription.caseSensitive ?
+                        stringFilterDescription.compareValue :
+                        stringFilterDescription.compareValue.toLowerCase();
+            }
+        }
+    }
+
+    public boolean test(@Nullable String curString) {
+        boolean result;
+        if (this.missing || (curString == null))    //Look for missing values
```
mbudiu-vmw: this would be more efficient using a separate class for each case.

platform/src/main/java/org/hillview/table/filters/StringFilter.java
```diff
@@ -0,0 +1,51 @@
...
+            this.missing = true;
+        else {
+            this.missing = false;
+            if (stringFilterDescription.asRegEx) {
+                this.regEx = Pattern.compile(stringFilterDescription.compareValue,
+                        stringFilterDescription.caseSensitive ? 0 : Pattern.CASE_INSENSITIVE);
+            } else {
+                this.compareTo = stringFilterDescription.caseSensitive ?
+                        stringFilterDescription.compareValue :
+                        stringFilterDescription.compareValue.toLowerCase();
```
mbudiu-vmw: This seems to be unnecessary.

platform/src/main/java/org/hillview/sketches/FindSketch.java
```diff
@@ -96,26 +98,26 @@ public JsonElement toJsonTree() {
...
         this.recordOrder = recordOrder;
     }
 
     @Override
     public Result create(ITable data) {
         long count = 0;
-        IStringFilter filter = this.toFind.getFilter();
         IRowIterator rowIt = data.getRowIterator();
         Schema toCheck = this.recordOrder.toSchema();
+        IStringFilter stringFilter = new StringFilterFactory().getFilter(this.stringFilterDescription);
```
mbudiu-vmw: In general factories you do not allocate; they implement a singleton pattern.
So a factory should have a static field called instance which is statically allocated.

platform/src/main/java/org/hillview/table/filters/StringFilterDescription.java
```diff
@@ -1,65 +1,33 @@
-/*
```
mbudiu-vmw: I don't think you should delete the comments.

platform/src/main/java/org/hillview/table/filters/StringFilterFactory.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.commons.lang.StringUtils;
+import org.hillview.table.api.IStringFilter;
+
+import javax.annotation.Nullable;
+import java.util.regex.Pattern;
+
+public class StringFilterFactory {
+
+    public IStringFilter getFilter(StringFilterDescription stringFilterDescription) {
+        assert(stringFilterDescription != null);
```
mbudiu-vmw: you don't need this assertion if the argument is not Nullable

platform/src/main/java/org/hillview/table/filters/StringFilterFactory.java
```diff
@@ -0,0 +1,90 @@
+package org.hillview.table.filters;
```
mbudiu-vmw: you need to add a copyright

platform/src/main/java/org/hillview/table/filters/StringRowFilterDescription.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+package org.hillview.table.filters;
+
+import org.hillview.table.api.*;
+
+import javax.annotation.Nullable;
+
+public class StringRowFilterDescription implements ITableFilterDescription {
+    @Nullable
```
mbudiu-vmw: I don't think this can be null, and neither is the next one.

platform/src/main/java/org/hillview/table/filters/StringFilterFactory.java
```diff
@@ -0,0 +1,90 @@
...
+
+import org.apache.commons.lang.StringUtils;
+import org.hillview.table.api.IStringFilter;
+
+import javax.annotation.Nullable;
+import java.util.regex.Pattern;
+
+public class StringFilterFactory {
+
+    public IStringFilter getFilter(StringFilterDescription stringFilterDescription) {
```
mbudiu-vmw: the constructor should be private in a factory with a singleton.

platform/src/test/java/org/hillview/test/ConfigGrammarTest.java
```diff
@@ -27,6 +27,7 @@
  * Test the parser for the cluster config grammar.
  */
 public class ConfigGrammarTest extends BaseTest {
+    /*
```
mbudiu-vmw: please uncomment this.

platform/src/test/java/org/hillview/test/TrellisShapeTest.java
```diff
@@ -11,15 +11,17 @@ public void tShapeTest() {
...
+            if (toPrint)
                 System.out.printf("\nAspect ratio %f\n", ratio);
-            }
             for (int n = 10; n <= 25; n++) {
                 ComputeTrellisShape.TrellisShape optShape = tShape.getShape(n);
-                if (optShape.x_num * optShape.y_num < n)
-                    System.out.println("Constraints are not feasible for ");
-                System.out.printf("%d: %d * %d, Size %.0f * %.0f, Coverage: %f \n", n, optShape.x_num,
-                        optShape.y_num, optShape.x_len, optShape.y_len, optShape.coverage);
+                if(toPrint) {
```
mbudiu-vmw: space after if

platform/src/main/java/org/hillview/sketches/FindSketch.java
parikg: Perhaps then factory is the wrong term for this class? It needs to allocate a different type of filter based on the stringFilterDescription it receives. So I don't see how to do that statically.

parikg:
So the StringFilterDescription is not passed to the constructor, it is only
given to the getfilter method, is that correct?

P.

mbudiu-vmw:
Yes

parikg:
Ok. I think it is fixed now.

platform/src/main/java/org/hillview/table/filters/StringFilterFactory.java
```diff
@@ -0,0 +1,112 @@
...
+package org.hillview.table.filters;
+
+import org.apache.commons.lang.StringUtils;
+import org.hillview.table.api.IStringFilter;
+
+import javax.annotation.Nullable;
+import java.util.regex.Pattern;
+
+public class StringFilterFactory {
+    public static IStringFilter instance = null;
```
mbudiu-vmw: Initialize this with `new StringFilterFactory()`.

mbudiu-vmw: The instance is of type StringFilterFactory, not IStringFilter.

platform/src/main/java/org/hillview/table/filters/StringFilterFactory.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public IStringFilter getFilter(StringFilterDescription stringFilterDescription) {
```
mbudiu-vmw: This function was fine previously. You should revert to the previous implementation.

platform/src/main/java/org/hillview/sketches/FindSketch.java
```diff
@@ -96,26 +98,28 @@ public JsonElement toJsonTree() {
...
         this.recordOrder = recordOrder;
     }
 
     @Override
     public Result create(ITable data) {
         long count = 0;
-        IStringFilter filter = this.toFind.getFilter();
         IRowIterator rowIt = data.getRowIterator();
         Schema toCheck = this.recordOrder.toSchema();
+        IStringFilter stringFilter = new StringFilterFactory().
```
mbudiu-vmw: This should be `StringFilterFactory.instance.getFilter(...)`.

platform/src/main/java/org/hillview/table/filters/StringFilterFactory.java
mbudiu-vmw: Also. make the constructor of the StringFilterFactory private. No one else should be able to call it.

platform/src/main/java/org/hillview/table/filters/StringRowFilterDescription.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * This filter maps a given Table to a Table that only contains the given value in the
+     * specified column.
+     */
+    public class StringRowFilter implements ITableFilter {
+        private final IColumn column;
+        private final IStringFilter stringFilter;
+        StringRowFilter(ITable table) {
+            this.stringFilter = new StringFilterFactory().getFilter(stringFilterDescription);
```
mbudiu-vmw: Same here: `StringFilterFactory.instance.getFilter`

platform/src/test/java/org/hillview/test/StringFilterFactoryTest.java
```diff
@@ -18,45 +18,46 @@
...
-        filter = desc.getFilter();
+        IStringFilter filter = new StringFilterFactory().getFilter(desc);
         Assert.assertTrue(filter.test("bob"));
         Assert.assertTrue(filter.test("Bob"));
         Assert.assertFalse(filter.test("mike"));
 
-        desc = new StringFilterDescription("^bo.*", true, true, false);
-        filter = desc.getFilter();
+        desc = new StringFilterDescription("bob", false, false, true);
+        filter = new StringFilterFactory().getFilter(desc);
```
mbudiu-vmw: you will have to change all these to refer to the instance as well; the private constructor will tell you that.

platform/src/main/java/org/hillview/table/filters/StringFilterFactory.java
parikg: I am still a but puzzled about the role of instance. It seems that I could make getFilter() a static method of this class, without any need for instance.

mbudiu-vmw:
A static method is fine.

platform/src/main/java/org/hillview/table/filters/StringFilterFactory.java
```diff
@@ -0,0 +1,111 @@
...
+package org.hillview.table.filters;
+
+import org.apache.commons.lang.StringUtils;
+import org.hillview.table.api.IStringFilter;
+
+import javax.annotation.Nullable;
+import java.util.regex.Pattern;
+
+public class StringFilterFactory {
+    public static StringFilterFactory instance = new StringFilterFactory();
```
mbudiu-vmw: You are right, let's take this out.
Please add a private constructor: 
`private StringFilterFactory() {}`

mbudiu-vmw: So we don't need an instance at all.

platform/src/main/java/org/hillview/table/filters/StringFilterFactory.java
```diff
@@ -0,0 +1,111 @@
...
+
+    public IStringFilter getFilter(StringFilterDescription stringFilterDescription) {
```
mbudiu-vmw: instead just make this method static.

PR #393, opened 2018-09-06
Address the pylint warnings

SivaMaplelabs:
Pylint score = 10


mbudiu-vmw:
Rather than disable some of these warnings you should feel free to change the code instead to produce no warnings.

mbudiu-vmw:
The missing docstring may be difficult, but you can leave it in.

mbudiu-vmw:
You should use pylint3

PR #418, opened 2018-09-27
Ignore the empty lines instead of halting with 'Could not parse line'

SivaMaplelabs:
Ignore the empty lines.


mbudiu-vmw:
Have you seen empty lines in the hillview logs?
I don't see any reason why there should be any.

SivaMaplelabs:
@mbudiu-vmw By default empty lines will not be there in the hillview logs.  But if an user append an empty line/lines while checking the hillview logs,  this issue would occur. 
Thanks

PR #425, opened 2018-10-03
Integrate generic log driver to hillview

SivaMaplelabs:
Dear Hillview maintainers,

Please accept this PR.

**Analysis**
Currently, there is no way to analyse the general logs except Hillview system logs.

**Resolution**
Use java-grok package and develop generic log driver.

**Files changed**
    data/log/datanodelog
    data/log/hbaselog
    data/log/syslog
    data/log/yarnlog
    platform/pom.xml
    platform/src/main/java/org/hillview/storage/FileSetDescription.java
    platform/src/main/java/org/hillview/storage/GenericLogs.java
    platform/src/main/resources/patterns/README.md
    platform/src/main/resources/patterns/log-patterns 
    web/src/main/webapp/javaBridge.ts
    web/src/main/webapp/loadMenu.ts

**Tested**
    1. Syslogs
    2. Hadoop data node logs
    3. HBase logs
    4. YARN logs
    5. Linux auth logs


platform/src/main/resources/patterns/README.md
```diff
@@ -0,0 +1,44 @@
+**Grok**
```
mbudiu-vmw: Is this the right place to put this documentation? 
We should probably add this to the userManual.src instead, and create a new section in the "loading data" of that document that links to this section. In fact, you should document the new menu in the "loading data" section.

platform/src/main/resources/patterns/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+**Grok Basics**
+
+Grok works by combining text patterns into something that matches your logs
+
+The syntax for a grok pattern is %{SYNTAX:SEMANTIC}
```
mbudiu-vmw: I would replace SEMANTIC with NAME.

platform/src/main/resources/patterns/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+SYNTAX : Name of the pattern that will match your text. For example, 3.44 will be matched by the NUMBER pattern and 55.3.244.1 will be matched by the IP pattern
+
+SEMANTIC : the identifier you give to the piece of text being matched. For example, 3.44 could be the duration of an event, so you could call it simply duration. Further, a string 55.3.244.1 might identify the client making a request
```
mbudiu-vmw: NAME: an identifier naming the matched text.

platform/src/main/resources/patterns/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+For the above example, your grok pattern would look something like this:
+```
+%{NUMBER:duration} %{IP:client}
+```
+For more patterns : https://github.com/thekrakken/java-grok/blob/master/src/main/resources/patterns/patterns
+
+**Example**
+
+Syslog :
```
mbudiu-vmw: Given this line from a syslog log:

platform/src/main/resources/patterns/README.md
```diff
@@ -0,0 +1,44 @@
...
+```
+For more patterns : https://github.com/thekrakken/java-grok/blob/master/src/main/resources/patterns/patterns
+
+**Example**
+
+Syslog : 
+```
+Sep 17 06:55:14,123 pndademocloud-hadoop-dn-3 CRON[25907]: (CRON) info (No MTA installed, discarding output)
+```
+The pattern for this could be :
```
mbudiu-vmw: This line can be parsed with the following pattern:

platform/src/main/resources/patterns/README.md
```diff
@@ -0,0 +1,44 @@
...
+```
+%{SYSLOGTIMESTAMP:timestamp} (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:logsource} %{GREEDYDATA:message}
+```
+Grok returns the following event
```
mbudiu-vmw: This creates the following log record.

web/src/main/webapp/loadMenu.ts
```diff
@@ -322,6 +334,34 @@ class CSVFileDialog extends Dialog {
...
+
+class GenericLogDialog extends Dialog {
+    constructor() {
+        super("Load Generic Logs",
+            "Loads log files from all machines that are part of the service.");
+        this.addTextField("folder", "Folder", FieldKind.String, "/",
+            "Folder on the remote machines where all the log files are found.");
+        this.addTextField("fileNamePattern", "File name pattern", FieldKind.String, "*.log",
+            "Shell pattern that describes the names of the files to load.");
+        this.addTextField("logFormat", "Log format", FieldKind.String, "%{SYSLOG}",
```
mbudiu-vmw: Should we start with this being a dropdown box (addSelectField) with a small number of fixed patterns?
Do you mean for the user to type a full pattern here?

platform/src/main/resources/patterns/README.md
```diff
@@ -0,0 +1,44 @@
+**Grok**
+
+**Grok Basics**
```
mbudiu-vmw: We don't say what grok is. Perhaps this section should be entitled: "Specifying rules for parsing logs".
We can then say that we rely on the grok library for this purpose, and put the link to the grok documentation.

platform/src/main/resources/patterns/README.md
```diff
@@ -0,0 +1,44 @@
...
+timestamp : Sep 17 06:55:14.123
+```
+
+**Regular Expressions**
+
+Grok sits on top of regular expressions, so any regular expressions are valid in grok as well
+
+**Custom Patterns**
+
+Sometimes Hillview doesn’t have a pattern you need. For this, you have an option.
```
mbudiu-vmw: In the end you probably want to say that given a log pattern and a set of files, what you will get in Hillview is a table with columns corresponding to the names that you gave to the patterns.

platform/src/main/java/org/hillview/storage/GenericLogs.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.regex.Matcher;
+
+import io.krakens.grok.api.*;
+
+/**
+ * Reads Generic logs into ITable objects.
+ */
+
+public class GenericLogs {
+    private static Schema schema;
```
mbudiu-vmw: You cannot use static fields unless they never change; this code will be run on multiple threads and they will stomp on each other's static fields.

mbudiu-vmw: You don't need these fields to be static anyway.

platform/src/main/java/org/hillview/storage/GenericLogs.java
```diff
@@ -0,0 +1,113 @@
...
+    private static int schema_columns;
```
mbudiu-vmw: You don't seem to use schema_columns for anything useful, so I think you can just delete it. 
Instead you can check whether schema is not null, and get the column count from the schema.

platform/src/main/java/org/hillview/storage/GenericLogs.java
```diff
@@ -0,0 +1,113 @@
...
+    private static Grok grok;
+    private static String logFormat;
+
+    GenericLogs(String logFormat) {
+        this.logFormat = logFormat;
+        GrokCompiler grokCompiler = GrokCompiler.newInstance();
+        grokCompiler.registerDefaultPatterns();
+        grokCompiler.registerPatternFromClasspath("/patterns/log-patterns");
+        grok = grokCompiler.compile(this.logFormat, true);
```
mbudiu-vmw: You will have to use `this.grok`.

platform/src/main/java/org/hillview/storage/GenericLogs.java
```diff
@@ -0,0 +1,113 @@
...
+                for (Map.Entry<String,Object> entry : capture.entrySet()) {
+                    output[index] = entry.getValue().toString();
+                    index += 1;
+		}
+            }
+        }
+
+        @Override
+        public ITable load() {
+            try (BufferedReader reader = new BufferedReader(
```
mbudiu-vmw: Can't you just use this.getFileReader()? It handles a bunch of other cases, including compressed files and files that are not necessarily UTF-8 encoded.

platform/src/main/java/org/hillview/storage/GenericLogs.java
```diff
@@ -0,0 +1,113 @@
...
+                    new FileReader(this.filename))) {
+                if (schema_columns == 0) {
+                    String line = reader.readLine();
```
mbudiu-vmw: You have to handle the case when the log is empty.

platform/src/main/java/org/hillview/storage/GenericLogs.java
```diff
@@ -0,0 +1,113 @@
...
+                    Match gmatch = grok.match(line);
+                    final Map<String, Object> capture = gmatch.capture();
+                    for (Map.Entry<String,Object> entry : capture.entrySet()) {
+                        GenericLogs.schema.append(new ColumnDescription(entry.getKey(), ContentsKind.String));
+                    }
+                    this.columns = schema.createAppendableColumns();
+                    schema_columns = capture.size();
+                }
+                String[] fields = new String[schema_columns + 1];
```
mbudiu-vmw: This can be `new String[this.columns.length]`

platform/src/main/java/org/hillview/storage/GenericLogs.java
```diff
@@ -0,0 +1,113 @@
...
+    public static class LogFileLoader extends TextFileLoader {
+        LogFileLoader(final String path) {
+            super(path);
+        }
+
+        void parse(String line, String[] output) {
+            Match gm = grok.match(line);
+            final Map<String, Object> capture = gm.capture();
+            if (capture.size() > 0) {
+                output[0] = Utilities.getHostName();
```
mbudiu-vmw: You should probably save the hostname in a field - this function may be very expensive and it always returns the same thing.

web/src/main/webapp/loadMenu.ts
SivaMaplelabs: @mbudiu-vmw Will add the drop-down box as part of next PR.

SivaMaplelabs:
@mbudiu-vmw Have addressed all the comments except documentation.

platform/src/main/resources/patterns/README.md
SivaMaplelabs: @mbudiu-vmw Should I change userManual.md as well?  Looks like there are many line numbers references so need to change the entire file.

mbudiu-vmw:
User manual.md is generated by the numbersections.py script

mbudiu-vmw:
Out of office today, will review Monday

platform/src/main/java/org/hillview/storage/GenericLogs.java
```diff
@@ -0,0 +1,117 @@
...
+
+import io.krakens.grok.api.*;
+
+/**
+ * Reads Generic logs into ITable objects.
+ */
+public class GenericLogs {
+    private static String hostName;
+    private Schema schema;
+    private static Grok grok;
```
mbudiu-vmw: I don't think these two should be static.
If you open a second log it will overwrite the format of the first one.

SivaMaplelabs: @mbudiu-vmw Agree with you. Have changed those two to non-static. Thanks

platform/src/main/java/org/hillview/storage/GenericLogs.java
```diff
@@ -0,0 +1,117 @@
...
+        this.grok = grokCompiler.compile(this.logFormat, true);
+        schema = new Schema();
+    }
+
+    static {
+        GenericLogs.hostName = Utilities.getHostName();
+    }
+
+    public static class LogFileLoader extends TextFileLoader {
+        private static GenericLogs genLog;
```
mbudiu-vmw: I think this one has to be non-static too.
Alternatively, you can make the LogFileLoader to be a non-static inner class of GenericLogs.

SivaMaplelabs: @mbudiu-vmw Done.

SivaMaplelabs:
@mbudiu-vmw Sure. Will raise PR for unit tests.

PR #426, opened 2018-10-03
Replace single line java comments with multi-line style comments

SivaMaplelabs:


platform/src/main/java/org/hillview/dataset/api/IJson.java
```diff
@@ -28,9 +28,10 @@
 import java.lang.reflect.Type;
 import java.time.Instant;
 
-// Unfortunately this module introduces many circular dependencies, because it has
-// to register various type adaptors.
-
+/**
```
mbudiu-vmw: This is not a proper javadoc comment. a Javadoc comment should describe what the described object is doing, and not what the shortcomings are. This should stay as a regular comment.

platform/src/main/java/org/hillview/dataset/api/IMonoid.java
```diff
@@ -32,8 +32,11 @@
 
     @Nullable
     default R reduce(List<R> data) {
-        // This implementation avoids allocating a zero
-        // if the list is non-empty.
+
+        /**
```
mbudiu-vmw: Don't use JavaDoc comments for code, only for fields, classes, etc.

SivaMaplelabs: Sure

SivaMaplelabs:
@mbudiu-vmw Have addressed the comments.

PR #428, opened 2018-10-03
There should be some ACK if worker started

SivaMaplelabs:
Currently, there is an ACK after the tomcat started(Tomcat started). But there is no ACK after the worker started.

### DIFF
--- a/bin/start.py
+++ b/bin/start.py
@@ -29,6 +29,9 @@ def start_worker(config, rh):
         " -jar " + config.service_folder + \
         "/hillview/hillview-server-jar-with-dependencies.jar " + rh.host + ":" + \
         str(config.worker_port) + " >nohup.out 2>&1 &")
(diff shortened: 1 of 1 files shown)


SivaMaplelabs:
@mbudiu-vmw Ok. But I didn't see any process fails after this check succeeds. In status.py also checking the same thing. 
https://github.com/vmware/hillview/pull/428/files#diff-9ea75055348a152656288b52117e3f4cR32
https://github.com/vmware/hillview/blob/master/bin/status.py#L22

It's good to have some ACK after the worker starts.

mbudiu-vmw:
The redeploy script already runs status after start.
Otherwise probably the best thing is to run it by hand every time you start.
We are not really in the business of cluster management, ideally one should use whatever tools are available already on the cluster where hillview is deployed (zookeeper?). These scripts are very rudimentary and don't really cover all cases needed in production. In particular, once the service fails we have no alerting or automatic restart.

SivaMaplelabs:
@mbudiu-vmw Sure.

PR #430, opened 2018-10-03
Handle file not found error

SivaMaplelabs:
root@hillview-root:~/DEPLOY-HILLVIEW/hillview/bin# python3 status.py config.json
Namespace(config='config.json')
Importing configuration from config.json
Traceback (most recent call last):
  File "status.py", line 49, in <module>
    main()
  File "status.py", line 44, in main
    config = ClusterConfiguration(args.config)
  File "/root/DEPLOY-HILLVIEW/hillview/bin/hillviewCommon.py", line 109, in __init__
    with open(file) as contents:
FileNotFoundError: [Errno 2] No such file or directory: 'config.json'

### DIFF
--- a/bin/status.py
+++ b/bin/status.py
@@ -6,6 +6,8 @@
 
 from argparse import ArgumentParser
 from hillviewCommon import ClusterConfiguration, RemoteHost
+import sys
+from os import path
 
 def check_webserver(config):
     """Checks if the Hillview web server is running"""
@@ -34,6 +36,13 @@ def main():
     parser = ArgumentParser()
     parser.add_argument("config", help="json cluster configuration file")
     args = parser.parse_args()
+    if 'config.json' not in args.config:
+        print("Please provide config.json")
+        print("Run : python3 status.py config.json")
+        sys.exit(1)
+    if not path.exists(args.config):
+        print("File " + args.config + " does not exists")
+        sys.exit(1)
     config = ClusterConfiguration(args.config)
     check_webserver(config)
     check_workers(config)


bin/status.py
```diff
@@ -34,6 +36,13 @@ def main():
     parser = ArgumentParser()
     parser.add_argument("config", help="json cluster configuration file")
     args = parser.parse_args()
+    if 'config.json' not in args.config:
+        print("Please provide config.json")
+        print("Run : python3 status.py config.json")
+        sys.exit(1)
+    if not path.exists(args.config):
```
mbudiu-vmw: We should handle this error in the constructor of ClusterConfiguration, then it will work for all files.
However, that method currently throws the right exception: No such file or directory. So the error message looks pretty clear to me.

bin/status.py
```diff
@@ -34,6 +36,13 @@ def main():
     parser = ArgumentParser()
     parser.add_argument("config", help="json cluster configuration file")
     args = parser.parse_args()
+    if 'config.json' not in args.config:
```
mbudiu-vmw: The file does not need to be named config.json. It can have any name.

bin/status.py
```diff
@@ -34,6 +36,13 @@ def main():
...
+        print("Please provide config.json")
+        print("Run : python3 status.py config.json")
```
mbudiu-vmw: The usage message should be printed by the argument parser, it already knows what the arguments are.

mbudiu-vmw:
I will close this PR, since I submitted a different set of fixes for this problem.

PR #452, opened 2018-10-11
Add bookmarked data menu to user manual

SivaMaplelabs:


docs/userManual.md
```diff
@@ -298,7 +302,18 @@ is deployed*.
...
+#### 2.3.4. Reading bookmarked data files
+
+Hillview can read the bookmarked data. The
+following menu allows the users to specify the files to load.
+
+![Specifying bookmarked files](bookmarked-data-menu.png)
+
+* File: A file containing the bookmarked data to load.
+
+* Tab lavel: A name to display for dataset.
```
mbudiu-vmw: typo

mbudiu-vmw:
We are planning to make this feature easier to use by saving the bookmarks file on the cloud side and returning a url. We should also probably rename this menu, because it's not really clear what it means.

docs/userManual.md
SivaMaplelabs: @mbudiu-vmw Done.

PR #453, opened 2018-10-12
Fixes in generic logs parsing

mbudiu-vmw:
@SivaMaplelabs : can you please review this code?


platform/src/main/java/org/hillview/storage/GenericLogs.java
```diff
@@ -80,25 +68,28 @@ void parse(String line, String[] output) {
         public ITable load() {
             try (BufferedReader reader = new BufferedReader(
                     this.getFileReader())) {
```
SivaMaplelabs: @mbudiu-vmw Please add the removed code to handle the empty file.
Else will face the NullPointerException.
0ccb64ea-9e84-4250-b0c1-b49d43335e32.loadTable() java.lang.Exception: java.lang.NullPointerException 
at org.hillview.dataset.LocalDataSet.lambda$map$1(LocalDataSet.java:118) 
Caused by: java.lang.NullPointerException

SivaMaplelabs:
@mbudiu-vmw Could you please handle the case that if the user provides wrong log pattern(for the Syslog %{HADOOP}) then schema will contain only host field as the condition(https://github.com/vmware/hillview/blob/b6c06902021c1ca3630dfccbe4dfb828aa7cc44b/platform/src/main/java/org/hillview/storage/GenericLogs.java#L83) will not match.

I missed this case to handle.

match

platform/src/main/java/org/hillview/storage/FileSetDescription.java
```diff
@@ -129,9 +129,11 @@ public ITable load() {
                     loader = new HillviewLogs.LogFileLoader(this.pathname);
                     break;
                 case "genericlog":
-                    GenericLogs genLog = new GenericLogs(FileSetDescription.this.getLogFormat());
-                    loader = new GenericLogs.LogFileLoader(this.pathname, genLog);
-                    break; 
+                    String format = FileSetDescription.this.getLogFormat();
+                    assert format != null;
```
SivaMaplelabs: @mbudiu-vmw I think the best way to handle this is to make the "logFormat" field mandatory in the UI as if the user does not provide anything in the logFormat it would halt while loading.
assert

platform/src/main/java/org/hillview/storage/GenericLogs.java
mbudiu-vmw: I will fix this, but if a log file is empty you will have other problems when you load data from multiple machines. The schema will not match between the various machines because empty files will have no columns. What we need to do is to create a schema even if there are no lines - the schema will need to be created directly from the pattern instead of being created from the first row.

platform/src/main/java/org/hillview/storage/FileSetDescription.java
mbudiu-vmw: We cannot easily validate the logformat in the UI.
Here we just assert it is not null - that makes the style checker from IntelliJ happy. 
Otherwise, because of the way we use the `@Nullable` annotation, we would have to initialize the corresponding field of the log format with some value, even if that value is never used (the format is populated by reflection when loading form JSON).

mbudiu-vmw:
What are the patterns you are using to load the 4 log files you have checked-in?

SivaMaplelabs:
1. Syslog
2. Datanode
3. Hbase
4. Yarn
https://github.com/vmware/hillview/blob/master/platform/src/main/resources/patterns/log-patterns

platform/src/main/java/org/hillview/storage/FileSetDescription.java
SivaMaplelabs: We can make the required fields mandatory so that there will not any null values for the fields.
https://www.w3schools.com/tags/att_input_required.asp

SivaMaplelabs: We can fix this as part of https://github.com/vmware/hillview/issues/456.

mbudiu-vmw:
@SivaMaplelabs : I appologize for the big change; most of it is just moving the tests to different folders, since it was getting difficult to navigate the folder with tests. 
I have created several unit tests for the GenericLogs which load the files you provided.
Please take another look. I have integrated your changes already.
I am using a StringBuilder to construct the log lines; it is much more GC-efficient than appending strings.

mbudiu-vmw:
You will also notice that I created a special kind of column that has all values identical. There is no reason to store a copy of the hostname for every row, we store them in this fake column instead. The downside is that the column has to be added after parsing. But the storage cost is essentially free.

SivaMaplelabs:
@mbudiu-vmw Please merge this PR.

PR #460, opened 2018-10-12
Add shebang in lib.sh

SivaMaplelabs:
SC2148

### DIFF
--- a/bin/lib.sh
+++ b/bin/lib.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 # Part of the Hillview configuration
 # This is a shell script which only defines some variables used by other scripts.
 


mbudiu-vmw:
The reason this does not have a shebang is that it should never be executed by itself.

SivaMaplelabs:
@mbudiu-vmw If a shell script is running inside another shell script also shebang is required.
Ref:
https://github.com/pndaproject/red-pnda/blob/develop/scripts/utils.sh#L1
https://github.com/pndaproject/red-pnda/blob/4bfa2001a9aa348d4a114f24f91fbce655ead361/scripts/install-dependencies.sh#L5

Issue #378, opened 2018-07-24
Upload script does not update files


lalithsuresh:
Running: ./upload-data.py config-onecloud.py -d <data> -s schema

The command runs and says it's doing rsync, but it doesn't update the schema file on the servers.

SivaMaplelabs:
https://github.com/vmware/hillview/pull/466

SivaMaplelabs:
@mbudiu-vmw Can close this issue.

mbudiu-vmw:
I think you have enough permissions to close issues as well, so feel free to do it when you think it should be done.

SivaMaplelabs:
@mbudiu-vmw Didn't close the fixed issues as I do not have the privilege to add the label as "fixed".

SivaMaplelabs:
Sure Mihai.

SivaMaplelabs:
@mbudiu-vmw I have the privilege to close the issues, which are opened by me only not by others.

mbudiu-vmw:
I am not sure that the issue is fixed. 
The line that was copying the files was removed by mistake only recently. 
I believe that the issue is about rsync not always updating files that have changed.

mbudiu-vmw:
@SivaMaplelabs : I am actually not the owner of this repository, this is owned and administered by VMware IT. So I don't see how I can grant you additional right. For now let's continue to operate in this way, I don't think this will preclude you from contributing.

PR #466, opened 2018-10-14
Fix for the issue #378

SivaMaplelabs:


bin/upload-data.py
```diff
@@ -69,8 +71,14 @@ def main():
     if not os.path.isabs(folder):
         folder = os.path.join(config.service_folder, folder)
         print("Folder is relative, using", folder)
+    files += args.files
     if args.common is not None:
-        copy_everywhere(config, args.common, folder, copyOptions)
+        files.append(args.common)
```
mbudiu-vmw: Only args.common file is copied to all machines, the other files are copied in round-robin amongst all machines.

SivaMaplelabs: Yeah,  got it.  Will change.

SivaMaplelabs: @mbudiu-vmw Done.

bin/upload-data.py
```diff
@@ -71,6 +71,11 @@ def main():
         print("Folder is relative, using", folder)
     if args.common is not None:
         copy_everywhere(config, args.common, folder, copyOptions)
+    if args.files:
+        copy_files(config, folder, args.files, copyOptions)
+    else:
+        print("No files to upload to the machines provided in a Hillview configuration")
+        exit(1)
```
mbudiu-vmw: There is no need to exit with 1 - sometimes you only want to use the args.common option

mbudiu-vmw:
I also don't see how this fixes issue #378 
We should check whether that issue is indeed correct.
I don't know when rsync decides to update a file.

bin/upload-data.py
SivaMaplelabs: @mbudiu-vmw Done.

SivaMaplelabs:
Currently, this script is not uploading the files to the given machines in a round-robin fashion. Only the schema file is uploading to all the machines.

SivaMaplelabs:
Only copying the schema, not the files.
https://github.com/vmware/hillview/blob/master/bin/upload-data.py#L72

PR #479, opened 2018-11-06
Multiple directory as input

arvind008:
We can have comma seperated directories as input.
If the directory does not exist it will return empty list


mbudiu-vmw:
If you can rebase and address the comments then we can merge this PR.

PR #480, opened 2018-11-06
Adding File Names in the UI

arvind008:
Adding file names in the UI


SivaMaplelabs:
Adding a "fileName" attribute in the table will give the flexibility to filter with filename if there are multiple files for the analyzing logs. Also gives the view about log belongs to which file in the worker node.

mbudiu-vmw:
There's a small risk that this column name will clash another column name from the pattern.
But if we do not allow users to type new patters this should be fine.

PR #481, opened 2018-11-06
Multiple regex patterns

arvind008:
Currently their is comparison of file names with file pattern using  String's matches() function.
It would be good to use Pattern.matches for ease of use.

### DIFF
--- a/platform/src/main/java/org/hillview/utils/Utilities.java
+++ b/platform/src/main/java/org/hillview/utils/Utilities.java
@@ -54,6 +54,9 @@ public class Utilities {
                 case '*':
                     s.append(".*");
                     break;
+                case '?':
+                    s.append(".");	
+                    break;
                 case '\\':
                     s.append("\\");
                     s.append(c);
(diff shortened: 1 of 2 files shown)


platform/src/main/java/org/hillview/utils/Utilities.java
```diff
@@ -54,11 +54,6 @@ public static String wildcardToRegex(@Nullable String wildcard) {
                 case '*':
                     s.append(".*");
                     break;
-                case '?':
```
mbudiu-vmw: Why are you deleting this code?
The user types a *shell* pattern, and the Java code runs a regular expression. This code converts a shell pattern into a regular expression.

arvind008: @mbudiu-vmw The shell pattern is interpreted as regular expression so conversion was not required. 

dot
question_mark

mbudiu-vmw: it is a different regular expression.
In the shell ? means the same as dot: one arbitrary character.
As a regex it is a modifier meaning: the previous regex is optional.

mbudiu-vmw:
I will close this PR for now. Note that we are still using file patterns and not regular expressions.
If you really want regular expressions you could add a checkbox to the input form allowing people to specify both regular expressions and filename patterns. Note that currently we only interpret patterns within the filename, but not within the directory name.

PR #482, opened 2018-11-10
Time range logs

mbudiu-vmw:
This commit reworks the generic log parsing. It adds UI and back-end support for reading only lines within a specified time range.

We expect that any log pattern the user specifies has a column named "Timestamp". That is used to create the timestamp, and to detect multi-line messages. For now the program will refuse to load a log that does not have a timestamp on the first line: this suggests strongly that the pattern is incorrect.

The loaded data will have a column called ParsingErrors; this column will contain all log lines that did not match the pattern. In this way we make sure we do not lose data silently.

The log patterns for parsing dates in Hadoop were wrong, so they were changed.

@SivaMaplelabs @arvind008 : please take a look; your other changes should be made after we merge this PR. Note that I have simplified the UI and now the path and file pattern are supplied in a single string.

Please ignore the columnResizer files; they should not have been part of this pull request.


platform/src/main/java/org/hillview/maps/FindFilesMapper.java
```diff
@@ -53,46 +52,40 @@ public FindFilesMapper(FileSetDescription description) {
...
-            fileNames = fileNames.limit(this.maxCount);
-        */
-        List<String> list = fileNames.collect(Collectors.toList());
-        String allNames = String.join(",", list);
-        HillviewLogger.instance.info("Files found", "{0}: {1}", list.size(), allNames);
+
         List<IFileReference> result = new ArrayList<IFileReference>();
-        for (String n: list)
+        for (String n: names)
+
```
SivaMaplelabs: Please remove this unnecessary new line.

arvind008:
@mbudiu-vmw i faced issue while showing the timestamp for the file shown below. I was testing for the line which mismatched the pattern, i had got the following error while showing the TimeStamp column.

File i was testing for
[test.txt](https://github.com/vmware/hillview/files/2571715/test.txt)

Error 
image

mbudiu-vmw:
I hope I have addressed the comments.
There are new columns too: directory and line number.
More robust handling of incorrect patterns.

PR #484, opened 2018-11-13
Adding Sample Hadoop Logs pattern and formats for reference

arvind008:
@mbudiu-vmw  Added File for reference to Hadoop Logs, it contains the paths,  patterns and sample log output.
After your PR for TimeRangeLogs is merged. Should we look to add in multiple file name pattern as input (comma separated).

for example:
    /var/log/hadoop-yarn/yarn/yarn*.log, /var/log/hadoop/hdfs/hadoop*.log

all the files within these 2 directories which matches the regex, has HADOOP as its logFormat.
So the output would contain logs from different directories.


mbudiu-vmw:
Please add a unit test that loads this file, probably in GenericLogTest. You can follow the patterns that appear there already.

platform/src/test/java/org/hillview/test/storage/GenericLogsTest.java
```diff
@@ -149,14 +148,14 @@ public void testWrongpattern() {
...
-        Assert.assertEquals("Table[9x42]", table.toString());
+        Assert.assertEquals("Table[8x42]", table.toString());
         IColumn col = table.getLoadedColumn(GenericLogs.parseErrorColumn);
         for (int i = 0; i < col.sizeInRows(); i++) {
             String s = col.getString(i);
             Assert.assertNotNull(s);
         }
     }
-
+
```
mbudiu-vmw: please configure your editor to delete spaces at the end of lines.

PR #501, opened 2018-12-22
Log file viewer - first working implementation

mbudiu-vmw:
Adds a very compact log viewer which can display a single remote file. Most functionality not yet available.

image


mbudiu-vmw:
@SivaMaplelabs: you may find this line of work interesting, it is designed to support a much better user experience when browsing logs. There is still a lot to implement, but this is a start.

PR #502, opened 2019-01-02
Adding curl to the dependencies

CMADHANKUMAR:


SivaMaplelabs:
@mbudiu-vmw cURL is not installed by default in some of the Linux distributions. Install-dependencies script is halting with "**The program 'curl' is currently not installed.  You can install it by typing:
sudo apt-get install curl**" error. This PR installs cURL if not installed already.

SivaMaplelabs:
@mbudiu-vmw Following are the reasons to check whether it exists
1. In some Linux distributions cURL is installed by default(No need to install again)
2. To avoid installing cURL if the user runs Install-dependencies script multiple times

It would be good to have the check. Please let me know your views on the same.

mbudiu-vmw:
We have lots of other more important problems to solve to focus on optimizing these scripts. So if it works it's fine.

SivaMaplelabs:
@mbudiu-vmw Addressed your comment. Please merge this PR.

PR #524, opened 2019-07-03
Fixed installation script

jiten-thakkar:


bin/install-dependencies.sh
```diff
@@ -18,7 +18,7 @@ case "$OSTYPE" in
         ${SUDO} ${INSTALL} install curl -y
 	curl -sL https://deb.nodesource.com/setup_11.x | ${SUDO} -E bash -
 esac
-${SUDO} ${INSTALL} wget install maven ${NODEJS} ${NPM} ${LIBFORTRAN} unzip gzip
+${SUDO} ${INSTALL}  install wget maven ${NODEJS} ${NPM} ${LIBFORTRAN} unzip gzip
```
mbudiu-vmw: I wonder if the second install is necessary at all...

bin/lib.sh
```diff
@@ -11,7 +11,7 @@ TOMCATVERSION="9.0.4"
 
 case "$OSTYPE" in
     linux*)
-        LIBFORTRAN="ligbfortran3"
+        LIBFORTRAN="libgfortran3"
```
mbudiu-vmw: this explains why this was failing...

bin/install-dependencies.sh
mbudiu-vmw: Actually yes, it is, of course.

PR #526, opened 2019-07-11
Batch log analysis

huilyu2:


platform/src/main/java/org/hillview/main/BatchLogAnalysis.java
```diff
@@ -0,0 +1,188 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.commons.cli.*;
+
+
+public class BatchLogAnalysis {
```
mbudiu-vmw: You should add a comment describing what this class is supposed to do.

platform/src/main/java/org/hillview/main/BatchLogAnalysis.java
```diff
@@ -0,0 +1,188 @@
...
+
+    private static class HeatmapData {
+        long[][] matrix;
+        JsonList<String> errorCodeLabels = new JsonList<>();
+        ArrayList<String> timeLabels = new ArrayList<>();
+    }
+
+    private static HeatmapData heatmapErrTime(FileSetDescription desc, LocalDataSet<Empty> local, int numOfTimestampBuckets) {
```
mbudiu-vmw: You should also document this function, even if it is private.
You should also add comments before each group of statements describing what they are doing. This will make reading the code easier.

platform/src/main/java/org/hillview/main/BatchLogAnalysis.java
mbudiu-vmw: The comments should document the expected input structure: logs that are in the syslog rfc5424 format, and that have a field called "errorCode" in the "structured data" column.

platform/src/main/java/org/hillview/main/BatchLogAnalysis.java
```diff
@@ -0,0 +1,188 @@
...
+        for (int x = 0; x < numOfTimestampBuckets; x++){
+            double time = dataRange.min + x * (dataRange.max - dataRange.min) / numOfTimestampBuckets;
+            Instant instantTime = Converters.toDate(time);
+            String stringDate = Converters.toString(instantTime);
+            heatmapData.timeLabels.add(stringDate);
+        }
+        return heatmapData;
+    }
+
+    private static void saveHeatmapToFile(HeatmapData heatmapData, String filepath) {
```
mbudiu-vmw: For this function you should document the file format you are using. If this is a data frame, just say that.

platform/src/main/java/org/hillview/main/BatchLogAnalysis.java
```diff
@@ -0,0 +1,188 @@
...
+                        sb.append(heatmapData.errorCodeLabels.get(y).replaceAll(",",""));
+                        sb.append(',');
+                        sb.append(heatmapData.matrix[x][y]);
+                        sb.append('\n');
+                    }
+                }
+            }
+            writer.write(sb.toString());
+        } catch (FileNotFoundException e) {
+            System.out.println(e.getMessage());
```
mbudiu-vmw: It;s usually not a good idea to completely swallow an exception.
Why don't you just stop processing in this case?

platform/src/main/java/org/hillview/main/BatchLogAnalysis.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    public static void getBugHeatmaps(FileSetDescription desc, LocalDataSet<Empty> local, String logDir, String figDir, int numOfTimestampBuckets) {
```
mbudiu-vmw: Please add a JavaDoc comment to this method describing each of the arguments.
I don't think you need the local argument, you can just create it locally.

PR #527, opened 2019-07-17
First cut of private histogram data structure

pratiksha:


platform/src/main/java/org/hillview/sketches/DyadicHistogramBuckets.java
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright (c) 2018 VMware Inc. All Rights Reserved.
```
mbudiu-vmw: 2019

platform/src/main/java/org/hillview/sketches/DyadicHistogramBuckets.java
```diff
@@ -0,0 +1,118 @@
...
+    private double maxValue;
+    private int numOfBuckets;
+    private final double bucketSize;
+    private final double range;
+
+    private final int granularity;
+    private final int numLeaves;
+
+    private void roundMinMax() {
+        long intMin = (long)Math.floor(this.minValue);
```
mbudiu-vmw: I don't think this will work for arbitrary double values for min and max.
Consider the case where min = 0.1 and max is 0.2.

platform/src/main/java/org/hillview/sketches/DyadicHistogramBuckets.java
```diff
@@ -0,0 +1,118 @@
...
+        long intMax = (long)Math.ceil(this.maxValue);
+
+        long minRem = intMin % this.granularity;
+        long maxRem = intMax % this.granularity;
+
+        intMin = intMin - minRem;
+        if ( maxRem > 0 ) { // Don't accidentally add an extra leaf!
```
mbudiu-vmw: no spaces around expression

platform/src/main/java/org/hillview/sketches/DyadicHistogramBuckets.java
```diff
@@ -0,0 +1,118 @@
...
+        this.bucketSize = (this.maxValue - this.minValue) / this.numOfBuckets;
+        this.numLeaves = (int)((this.maxValue - this.minValue) / this.granularity);
+
+        // Preserves semantics, will make noise computation easier
+        if ( this.numLeaves < this.numOfBuckets ) {
+            this.numOfBuckets = this.numLeaves;
+        }
+    }
+
+    public int indexOf(double value) {
```
mbudiu-vmw: This code is very similar to DoubleHistogramBuckets.
Can you inherit from that class?

mbudiu-vmw: Actually, I think that probably this is not how this should be done; it is simpler to first round each value in the dataset down to the closest discretization boundary and then count it. @parikg can explain what I mean if this is unclear.

platform/src/main/java/org/hillview/sketches/Histogram.java
```diff
@@ -32,6 +33,35 @@ public Histogram(final IHistogramBuckets bucketDescription) {
         this.buckets = new long[bucketDescription.getNumOfBuckets()];
     }
 
+    // Note that we ceil in order to round up to the nearest leaf
```
mbudiu-vmw: I think you should probably extend the Histogram class with a PrivateHistogram class and add these methods there.

platform/src/main/java/org/hillview/sketches/Histogram.java
```diff
@@ -32,6 +33,35 @@ public Histogram(final IHistogramBuckets bucketDescription) {
...
+    public static int intLog2(int x) {
```
mbudiu-vmw: Hopefully x is not 0.

platform/src/main/java/org/hillview/sketches/Histogram.java
```diff
@@ -32,6 +33,35 @@ public Histogram(final IHistogramBuckets bucketDescription) {
...
+        return (int)(Math.ceil(Math.log(x) / Math.log(2)));
+    }
+
+    // The amount of noise depends on the number of "leaves" that are required to compute the interval.
+    public int noiseMultiplier(int bucketIdx) {
+        // Cast down - the noise operation is only valid on the dyadic histogram leaves
```
mbudiu-vmw: If you subclass Histogram then this can be made slightly safer too by storing a pointer already cast.

platform/src/main/java/org/hillview/sketches/Histogram.java
```diff
@@ -32,6 +33,35 @@ public Histogram(final IHistogramBuckets bucketDescription) {
...
+    /* Add Laplace noise compatible with the binary mechanism to each bucket.
+     * This operation cannot be undone.
+     * Note that the noise scale provided should be log T / epsilon where T is
+     * the total number of leaves in the unfiltered histogram.
+     * This is not computed in the function, because the range of the buckets might be
+     * smaller than the full range T if the buckets are from a filtered histogram. */
+    public void addDyadicLaplaceNoise(double scale) {
+        LaplaceDistribution dist = new LaplaceDistribution(0, scale); // TODO: (more) secure PRG
+        for ( int i = 0; i < this.buckets.length; i++ ) {
+            this.buckets[i] += dist.sample() * noiseMultiplier(i);
```
mbudiu-vmw: The style is this.noiseMultiplier(i).

platform/src/main/java/org/hillview/sketches/Histogram.java
```diff
@@ -32,6 +33,35 @@ public Histogram(final IHistogramBuckets bucketDescription) {
...
+    }
+
+    /* Add Laplace noise compatible with the binary mechanism to each bucket.
+     * This operation cannot be undone.
+     * Note that the noise scale provided should be log T / epsilon where T is
+     * the total number of leaves in the unfiltered histogram.
+     * This is not computed in the function, because the range of the buckets might be
+     * smaller than the full range T if the buckets are from a filtered histogram. */
+    public void addDyadicLaplaceNoise(double scale) {
+        LaplaceDistribution dist = new LaplaceDistribution(0, scale); // TODO: (more) secure PRG
```
mbudiu-vmw: Please work with Udi on the design of the API of such a class.

platform/src/test/java/org/hillview/test/dataStructures/PrivateHistogramTest.java
```diff
@@ -0,0 +1,138 @@
...
+        final int granularity = 10; // but leaves of size 10
+        DyadicHistogramBuckets buckDes = new DyadicHistogramBuckets(min, max, numBuckets, granularity);
+        Histogram hist = new Histogram(buckDes);
+
+        DoubleArrayColumn col = generateLinearColumn(min, max, 1);
+        FullMembershipSet fMap = new FullMembershipSet(col.sizeInRows());
+        hist.create(col, fMap, 1.0, 0, false);
+
+        // manually compute expected multiplier
+        for ( int i = 0; i < hist.getNumOfBuckets(); i++ ) {
```
mbudiu-vmw: no spaces

platform/src/main/java/org/hillview/sketches/DyadicHistogramBuckets.java
pratiksha: Hmm, these functions will certainly incur floating-point errors if we want to allow non-integer granularity/range, which is dangerous if it leads us to add the wrong amount of noise to a bucket. It's not clear to me how to implement `numLeavesInBucket` safely if `indexOf` is implemented as you suggested. One solution could be to switch to BigDecimal, but I don't think that entirely saves us from precision errors. Let me know if you have other suggestions.

Relatedly, we'll eventually have to implement floating-point-secure Laplace sampling: https://crysp.uwaterloo.ca/courses/pet/F18/cache/Mironov.pdf

platform/src/main/java/org/hillview/sketches/Histogram.java
pratiksha: If I subclass Histogram, it seems like I can't use the HistogramSketch directly, as it returns a Histogram rather than the subclass (and slices away the relevant postprocessing method and attributes). Creating a new PrivateHistogramSketch, etc., seems like it will involve duplicating a lot of code. Any suggestions on a better way to do this?

platform/src/main/java/org/hillview/sketches/Histogram.java
```diff
@@ -32,6 +33,38 @@ public Histogram(final IHistogramBuckets bucketDescription) {
...
+        System.out.println("Buckets: " + this.buckets.length);
```
mbudiu-vmw: You probably want to take out the printlns.

platform/src/main/java/org/hillview/sketches/DyadicHistogramBuckets.java
```diff
@@ -0,0 +1,183 @@
...
+import org.hillview.table.api.IColumn;
+
+import java.util.Arrays;
+
+/**
+ * DyadicHistogramBuckets ensures that every bucket is a multiple of the leaf granularity.
+ * It is built on data that can be converted to doubles.
+ * The last bucket is right-inclusive.
+ */
+public class DyadicHistogramBuckets implements IHistogramBuckets {
```
mbudiu-vmw: The comment will have to be much more elaborate; you can include a link to the paper with the binary mechanism.
You have to explain what the leaves are and what the rules are for creating the buckets.

platform/src/main/java/org/hillview/sketches/DyadicHistogramBuckets.java
```diff
@@ -0,0 +1,183 @@
...
+    private double minValue;
+    private double maxValue;
+    private int numOfBuckets;
+
+    private final double bucketSize;
+
+    private final double granularity;
```
mbudiu-vmw: You will also have to document these fields.

platform/src/main/java/org/hillview/sketches/Histogram.java
mbudiu-vmw: Do you still need these methods here?
I see that they are already in the PrivateHistogram.

platform/src/main/java/org/hillview/sketches/PrivateHistogram.java
```diff
@@ -0,0 +1,60 @@
...
+    private final DyadicHistogramBuckets dyadicBuckets;
+
+    private final double maxValue;
+    private final double minValue;
+
+    // Confidence intervals
+    private double[] confMins;
+    private double[] confMaxes;
+
+    public PrivateHistogram(final IHistogramBuckets bucketDescription) {
```
mbudiu-vmw: I think you should make the parameter be DyadicHistogramBuckets.

platform/src/main/java/org/hillview/sketches/PrivateHistogram.java
```diff
@@ -0,0 +1,60 @@
...
+        this.confMaxes = new double[dyadicBuckets.getNumOfBuckets()];
+    }
+
+    // Note that we ceil in order to round up to the nearest leaf
+    public static int intLog2(int x) {
+        assert (x > 0);
+        return (int)(Math.ceil(Math.log(x) / Math.log(2)));
+    }
+
+    // The amount of noise depends on the number of "leaves" that are required to compute the interval.
```
mbudiu-vmw: This comment should be inside the method, not about the method.

platform/src/main/java/org/hillview/sketches/PrivateHistogram.java
```diff
@@ -0,0 +1,60 @@
...
+
+        this.minValue = dyadicBuckets.getMin();
+        this.maxValue = dyadicBuckets.getMax();
+
+        // Not currently used
+        this.confMins = new double[dyadicBuckets.getNumOfBuckets()];
+        this.confMaxes = new double[dyadicBuckets.getNumOfBuckets()];
+    }
+
+    // Note that we ceil in order to round up to the nearest leaf
```
mbudiu-vmw: put the comment on the statement that makes the ceil.

platform/src/main/java/org/hillview/sketches/PrivateHistogram.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public PrivateHistogram(final IHistogramBuckets bucketDescription) {
+        super(bucketDescription);
+        dyadicBuckets = (DyadicHistogramBuckets)bucketDescription;
+
+        this.minValue = dyadicBuckets.getMin();
+        this.maxValue = dyadicBuckets.getMax();
+
+        // Not currently used
+        this.confMins = new double[dyadicBuckets.getNumOfBuckets()];
```
mbudiu-vmw: Do you need to store these or can you just compute them?

platform/src/main/java/org/hillview/sketches/PrivateHistogram.java
```diff
@@ -0,0 +1,60 @@
...
+    // The amount of noise depends on the number of "leaves" that are required to compute the interval.
+    public int noiseMultiplier(int bucketIdx) {
+        int nLeaves = dyadicBuckets.numLeavesInBucket(bucketIdx);
+        if (nLeaves == 1) return 1;
+
+        return intLog2(nLeaves);
+    }
+
+    /* Add Laplace noise compatible with the binary mechanism to each bucket.
+     * This operation cannot be undone.
```
mbudiu-vmw: This comment is not really necessary; even computing a histogram "cannot be undone".

web/src/main/java/org/hillview/targets/TableTarget.java
```diff
@@ -309,8 +309,49 @@ HistogramSketch getSketch() {
     @HillviewRpc
     public void histogram(RpcRequest request, RpcRequestContext context) {
         HistogramArgs info = request.parseArgs(HistogramArgs.class);
+    }
+
+    static class PrivateHistogramArgs {
```
mbudiu-vmw: We will probably have to work together to implement a PrivateTableTarget class.
That class will store per column some of these parameters: epsilon, granularity, min. max.
Some of this code will move there.

web/src/main/java/org/hillview/targets/TableTarget.java
```diff
@@ -309,8 +309,49 @@ HistogramSketch getSketch() {
...
+        ColumnDescription cd = new ColumnDescription();
+        double samplingRate = 1.0; // Fix to exact count
+        double epsilon;
+        int granularity;
+        long seed;
+
+        double min;
```
mbudiu-vmw: Which min and max are these? For the data or for the query?

web/src/main/webapp/dataViews/schemaView.ts
```diff
@@ -177,6 +177,12 @@ export class SchemaView extends TSViewBase {
             help: "Plot the data in the selected columns as a histogram.  Applies to one or two columns only. " +
             "The data cannot be of type String.",
         }, selectedCount >= 1 && selectedCount <= 2);
+        this.contextMenu.addItem({
+            text: "Private histogram",
+            action: () => this.histogramSelected(),
+            help: "Plot the data in the selected columns as a private histogram. " +
+                "Applies to one numeric column only.",
+        }, selectedCount === 1);
```
mbudiu-vmw: You can also probably check here that the column is convertible to numeric in the enable condition.

web/src/main/webapp/dataViews/tsViewBase.ts
```diff
@@ -185,6 +191,39 @@ export abstract class TSViewBase extends BigTableView {
...
+	const range: DataRange = {
+	    presentCount: -1,
+	    missingCount: -1,
+	    min: minValue,
+	    max: maxValue,
+	};
+
+	const rowCount: number = 1;
+	const rr = this.createPrivateHistogramRequest(args);
+        const axisData = new AxisData(cd, range);
```
mbudiu-vmw: inconsistent indentation

web/src/main/webapp/dataViews/tsViewBase.ts
```diff
@@ -185,6 +191,39 @@ export abstract class TSViewBase extends BigTableView {
...
+        const renderer = new PrivateHistRenderer(new PageTitle("Private histogram"), this.page,
```
mbudiu-vmw: you probably want to put the column name in the title.

web/src/main/webapp/dataViews/tsViewBase.ts
```diff
@@ -234,6 +273,15 @@ export abstract class TSViewBase extends BigTableView {
         }
     }
 
+    protected privateHistogram(columns: string[],
+			       minValue: number, maxValue: number,
+			       epsilon: number, granularity: number,
+			       bucketCount: number): void {
+        const cds = this.schema.getDescriptions(columns);
+	console.assert(cds.length === 1);
```
mbudiu-vmw: inconsistent indentation

web/src/main/webapp/dataViews/tsViewBase.ts
```diff
@@ -272,6 +320,56 @@ export abstract class TSViewBase extends BigTableView {
...
 
+    protected privateHistSelected(): void {
+        if (this.getSelectedColCount() !== 1) {
+            this.page.reportError("Must select 1 column for private histogram");
+            return;
+        }
+
+        let title = "Private Histogram";
+	const d = new Dialog(title, "Plot a private histogram with the specified granularity and privacy budget.");
+        const epsilon = d.addTextField("epsilon", "Privacy budget, epsilon", FieldKind.Double, "0.25",
```
mbudiu-vmw: As discussed, some of this will move from the menu to the server-side PrivateTableTarget.

web/src/main/webapp/javaBridge.ts
```diff
@@ -252,6 +252,34 @@ export interface HistogramArgs {
...
+
+export interface PrivateHistogramArgs {
+    cd: IColumnDescription;
+    seed: number;
+    samplingRate: number;
+    bucketCount: number;
+    epsilon: number;
+    granularity: number;
+
+    // only used when doing double histograms
```
mbudiu-vmw: The comment is no longer needed.

web/src/main/webapp/ui/ui.ts
```diff
@@ -63,7 +63,7 @@ export class HtmlString {
     }
 }
 
-export type ViewKind = "Table" | "Histogram" | "2DHistogram" | "Heatmap" |
+export type ViewKind = "Table" | "Histogram" | "PrivateHistogram" | "2DHistogram" | "Heatmap" |
```
mbudiu-vmw: Is this used anywhere?

pratiksha: Hm, I guess it's not used right now. Will we need this eventually? ViewKind is used in several places, not sure which of them we will need to adapt for the private histogram as well.

web/src/main/webapp/dataViews/tableView.ts
```diff
@@ -84,11 +85,17 @@ export class TableView extends TSViewBase implements IScrollTarget, OnNextK {
...
     protected strFilter: StringFilterDescription;
 
+    protected privacySchema: PrivacySchema;
+
     // The following elements are used for Find
     protected findBar: FindBar;
 
     public constructor(
-        remoteObjectId: RemoteObjectId, rowCount: number, schema: SchemaClass, page: FullPage) {
+        remoteObjectId: RemoteObjectId,
```
mbudiu-vmw: alignment

web/src/main/webapp/dataViews/tableView.ts
```diff
@@ -99,6 +106,8 @@ export class TableView extends TSViewBase implements IScrollTarget, OnNextK {
         this.topLevel.onkeydown = (e) => this.keyDown(e);
         this.strFilter = null;
 
+	this.privacySchema = privacySchema;
+
```
mbudiu-vmw: no need for so many empty line; one below is enough

web/src/main/webapp/dataViews/tableView.ts
```diff
@@ -492,6 +501,140 @@ export class TableView extends TSViewBase implements IScrollTarget, OnNextK {
         rr.invoke(new NextKReceiver(this.page, this, rr, false, this.order, null));
     }
 
+    private createContextMenu(cd: IColumnDescription,
+			      selectedCount: number,
+			      visible: boolean,
+			      isPrivate: boolean) {
+        this.contextMenu.clear();
+        if (visible) {
```
mbudiu-vmw: I guess this should not happen for private columns.

web/src/main/webapp/dataViews/tableView.ts
```diff
@@ -492,6 +501,140 @@ export class TableView extends TSViewBase implements IScrollTarget, OnNextK {
...
+                },
+                help: "Extract a value associated with a specific key.",
+            }, selectedCount === 1 &&
+				     this.isKVColumn(this.getSelectedColNames()[0]));
+        } else {
+            this.contextMenu.addItem({
+                text: "Private Histogram",
+                action: () => this.privateHistSelected(),
+                help: "Plot the data in the selected columns as a private histogram. " +
+                    "Applies to one numeric column only.",
```
mbudiu-vmw: Instead of this line add a check below selectedCount === 1 && kindIsNumeric(cd.kind).
(You may also want to add something about dates too...)

web/src/main/webapp/loadMenu.ts
```diff
@@ -49,6 +49,23 @@ export class LoadMenu extends RemoteObject implements IDataView {
 
         this.top = document.createElement("div");
         this.testDatasetsMenu = new SubMenu([
+            { text: "Gaussian (1 column, private, CSV)",
```
mbudiu-vmw: Add a `TODO: to delete` here. We don't want this permanently.

mbudiu-vmw:
I want to merge this after you make these changes, and we can continue from there.
Try to also rebase on master.

web/src/main/webapp/dataViews/tableView.ts
pratiksha: What do you mean by this?

platform/src/main/java/org/hillview/table/Schema.java
```diff
@@ -39,7 +39,7 @@
  * A schema is an ordering of the columns, plus a map from a column name to a column description.
  * Column names are case-sensitive.
  */
-public final class Schema implements Serializable, IJson {
+public class Schema implements Serializable, IJson {
```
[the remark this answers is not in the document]
pratiksha: Originally it was so that PrivacySchema could extend it, but now PrivacySchema is not a subclass so it can go back to being final.

platform/src/main/java/org/hillview/sketches/PrivateHistogram.java
pratiksha: I think we can just compute them.

web/src/main/webapp/dataViews/tableView.ts
mbudiu-vmw: If the diff I saw one of the arguments was indented with a different amount.

web/src/main/webapp/ui/histogramPlot.ts
```diff
@@ -85,40 +89,55 @@ export class HistogramPlot extends Plot {
...
+		.attr("y1", (d) => this.yScale(d[0] - d[1]))
+		.attr("x2", 0)
+		.attr("y2", (d) => this.yScale(d[0] + d[2]))
+		.attr("stroke-width", 1)
+		.attr("stroke", "black")
+		.attr("stroke-linecap", "round")
+		.attr("transform", () => `translate(${this.barWidth / 2}, 0)`);
+	}
+
+	if (!this.isPrivate) {
```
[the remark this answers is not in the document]
pratiksha: I thought we decided not to show them because it's not meaningful?

platform/src/main/java/org/hillview/sketches/DyadicHistogramBuckets.java
```diff
@@ -176,6 +210,82 @@ public int numLeavesInBucket(int bucketIdx) {
...
+    // Returns the noise and the total variance of the variables used to compute the noise.
+    public Pair<Double, Double> noiseForBucket(int bucketIdx, boolean cdf) {
+        ArrayList<Pair<Long, Long>> intervals = bucketDecomposition(bucketIdx, cdf);
+
+        double noise = 0;
+        double variance = 0;
+        double scale = Math.log((this.globalMax - this.globalMin) / (this.epsilon * this.granularity)) / Math.log(2);
+        for (Pair<Long, Long> x : intervals) {
+            LaplaceDistribution dist = new LaplaceDistribution(0, scale); // TODO: (more) secure PRG
+            dist.reseedRandomGenerator(x.hashCode()); // Each node's noise should be deterministic, based on node's ID
```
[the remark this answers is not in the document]
pratiksha: Yes, we should definitely implement our own PRG.

web/src/main/webapp/ui/histogramPlot.ts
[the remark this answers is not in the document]
pratiksha: (And also because the numbers overlap with the error bars)

platform/src/main/java/org/hillview/sketches/DyadicHistogramBuckets.java
```diff
@@ -29,29 +34,40 @@
...
+    private double globalMax;
+
     private int numOfBuckets;
 
     private final double bucketSize; // The desired bucket size specified by the user or UI.
 
     private final double granularity; // The quantization interval (leaf size).
-    private int numLeaves; // Total number of leaves.
+    private final double epsilon; // The privacy budget parameter for the histogram.
+    private long numLeaves; // Total number of leaves.
```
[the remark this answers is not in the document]
pratiksha: Hmm. For a 1D histogram probably not, but for a heat map, this seems entirely possible (sqrt(2 billion) ~= 40000). Not sure it's worth having different types for each.

platform/src/main/java/org/hillview/sketches/DyadicHistogramBuckets.java
```diff
@@ -176,6 +210,82 @@ public int numLeavesInBucket(int bucketIdx) {
                 this.bucketLeftLeaves[bucketIdx]);
     }
 
+    // Return the dyadic decomposition of this interval, as a list of <left boundary, right boundary> pairs
+    // for each interval in the decomposition. The decomposition assumes that the first leaf of the dyadic tree
+    // is at index 0.
+    public static ArrayList<Pair<Long, Long>> dyadicDecomposition(Long left, Long right) {
```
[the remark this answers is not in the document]
pratiksha: Pair doesn't support primitive types?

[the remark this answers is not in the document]
mbudiu-vmw: Java generics never work with primitive types.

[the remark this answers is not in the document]
pratiksha: Ah, you were asking about the arguments. Yes, those can be `long`.

platform/src/main/java/org/hillview/sketches/Histogram.java
```diff
@@ -52,6 +64,10 @@ public void add(IColumn column, int currRow) {
             int index = this.bucketDescription.indexOf(column, currRow);
             if (index >= 0)
                 this.buckets[index]++;
+
+            for (int i = 0; i < index+1; i++) {
```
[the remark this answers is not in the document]
pratiksha: Hmm. This was intended for backwards-compatibility so that the CDF for the partial result would be sensible, and then the front-end doesn't have to have two separate procedures for computing the CDF (integrating the histogram vs. using the CDF from the backend). If it's computed at the root, the partial rendering won't work anymore, I think. But it might be expensive, yes.

platform/src/main/java/org/hillview/sketches/HistogramBase.java
```diff
@@ -23,5 +23,10 @@
 
 public abstract class HistogramBase implements Serializable, IJson {
     protected long[] buckets = new long[0];
+    protected long[] cdfBuckets = new long[0];
     protected long missingData;
+
+    // Confidence interval bounds for buckets
+    protected double[] confMins = new double[0];
```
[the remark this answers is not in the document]
pratiksha: It is somewhat annoying to do this: histogramPlot has a `histogram` member on which the histogram is rendered, and having these fields in the histogram itself allows us to plot everything directly from members of the histogram, and with minimal change for public vs. private. Happy to discuss the design further though.

platform/src/main/java/org/hillview/sketches/PrivateHistogram.java
```diff
@@ -17,29 +14,34 @@ public PrivateHistogram(final Histogram histogram) {
...
         for (int i = 0; i < this.histogram.buckets.length; i++) {
-            this.histogram.buckets[i] += dist.sample() * this.noiseMultiplier(i);
+            Pair<Double, Double> noise = this.bucketDescription.noiseForBucket(i, false);
+            this.histogram.buckets[i] += noise.first;
+            System.out.println("Bucket " + i + ": " + this.histogram.buckets[i]);
+            // Postprocess so that no buckets are negative
             this.histogram.buckets[i] = Math.max(0, this.histogram.buckets[i]);
+
+            // also set confidence intervals for this noise level
+            this.histogram.confMins[i] = 2*Math.sqrt(noise.second);
```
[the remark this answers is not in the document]
pratiksha: See response on HistogramBase.

platform/src/main/java/org/hillview/sketches/HistogramBase.java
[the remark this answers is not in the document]
mbudiu-vmw: That's fine, you can send a HistogramWithConfidence to the UI.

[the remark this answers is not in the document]
pratiksha: So are you suggesting that we should send a HistogramWithConfidence for both the public and private histograms (although the public histogram does not have confidence intervals)?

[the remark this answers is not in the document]
mbudiu-vmw: yes

mbudiu-vmw:
You have picked up a bunch of commits which don't belong to this PR; perhaps you didn't rebase on master?

mbudiu-vmw:
If the code does not affect any existing functionality I will merge it and we'll do some more work on it.

mbudiu-vmw:
I have merged this after rebasing, so I will close this PR.
Please branch off master for any new work.

pratiksha:
Yes, I forgot to rebase before the push - thanks for merging.

PR #528, opened 2019-08-05
BatchLogAnalysis separate nsx_manager and nsx_edge

huilyu2:
Modify BatchLogAnalysis to get heatmap data from nsx_manager folder and nsx_edge folder separately

### DIFF
--- a/platform/src/main/java/org/hillview/main/BatchLogAnalysis.java
+++ b/platform/src/main/java/org/hillview/main/BatchLogAnalysis.java
@@ -7,9 +7,13 @@ import org.hillview.dataset.api.IMap;
 import org.hillview.maps.ExtractValueFromKeyMap;
 import org.hillview.maps.FindFilesMapper;
 import org.hillview.maps.LoadFilesMapper;
+import org.hillview.maps.ProjectMap;
 import org.hillview.sketches.*;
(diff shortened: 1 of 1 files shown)


platform/src/main/java/org/hillview/main/BatchLogAnalysis.java
```diff
@@ -97,48 +111,54 @@ private static HeatmapData heatmapErrTime(FileSetDescription desc, int numOfTime
...
+    public static void getBugHeatmaps(String logDir, String figDir, int numOfTimestampBuckets) {
         File path = new File(logDir);
         String[] bugIDs = path.list(new FilenameFilter() {    // each subDir in logDir corresponds to one bug
             @Override
             public boolean accept(File current, String name) {
                 return new File(current, name).isDirectory();
             }
         });
+        // create empty directory for nsx_manager and nsx_edge respectively
+        new File(figDir + "/nsx_manager_syslog/").mkdirs();
```
mbudiu-vmw: Do we know that this is actually useful?

huilyu2: Separating nsx_manager and nsx_edge should be useful. The number of distinct errorCodes in each one should be smaller.

PR #542, opened 2019-09-18
Support for converting in schema view

mbudiu-vmw:
This also fixes a few other bugs related to renamed columns.


mbudiu-vmw:
Fixes #541

PR #544, opened 2019-09-29
This is a rebased version of Pratiksha's dyadic branch

mbudiu-vmw:


pratiksha:
It seems like something may have gone wrong in the merge? See e.g. https://github.com/vmware/hillview/blob/2d316c97bc262d91ad10ce3dfdec36db7d3d4dd7/platform/src/main/java/org/hillview/table/rows/GuessSchema.java#L201

mbudiu-vmw:
I think I was missing a commit. That's what happens to late evening commits.
I have pushed directly to master, hopefully now it works.
I plan to take a deeper look at some of your code and reorganize some of it to work with the local database target - which I haven't merged yet.

mbudiu-vmw:
We should synchronize so that the merge will be less painful.

PR #557, opened 2019-10-16
Private database histograms and selections

mbudiu-vmw:


mbudiu-vmw:
@pratiksha : this implements histograms and selection for private databases.
Some code is identical for PrivateTables; it has been moved into DPWrapper (perhaps not the best place).
I still have to do heatmaps for both databases and private databases.

mbudiu-vmw:
Actually this is not correct, because it does not include the quantization yet. Another PR upcoming.

PR #561, opened 2019-10-26
Support for Sum and CountNonNull aggregates

mbudiu-vmw:
More aggregates will follow.
Aggregated columns are shown with a different background:


mbudiu-vmw:
image

PR #568, opened 2019-11-08
Bug fix for noise computation

pratiksha:
Also accuracy benchmark.

@mbudiu-vmw ptal.

### DIFF
--- a/web/src/test/java/org/hillview/HistogramAccuracyTest.java
+++ b/web/src/test/java/org/hillview/HistogramAccuracyTest.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright (c) 2019 VMware Inc. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


web/src/test/java/org/hillview/HistogramAccuracyTest.java
```diff
@@ -0,0 +1,123 @@
...
+            for (int right = left; right < hist.getBucketCount(); right++) {
+                Pair<Double, Double> noise = dd.noiseForRange(left, right, epsilon,
+                        dist, baseVariance, false);
+                sqtot += Math.pow(noise.first, 2);
+                abstot += Math.abs(noise.first);
+                n++;
+            }
+        }
+
+        System.out.println("Bucket count: " + hist.getBucketCount());
```
mbudiu-vmw: An assert is more useful than a print, but anything is better than nothing.

pratiksha: Ah, this is really a benchmark and not a regression test, but I guess accuracy should stay constant here, so I can write asserts instead.

PR #575, opened 2019-11-25
WIP secure sampling implementation

pratiksha:
Not quite ready to merge as I'm still looking at performance, but we can start the code review.


mbudiu-vmw:
I will commit some changes to your code.
Some of the are cleanups found by the code inspection, perhaps I should have avoided doing this.
Indeed, the noise computation is now much faster - it seems to be in the noise.
You should probably install the same dataset I have so we don't have to keep editing this program.

PR #591, opened 2020-01-11
Experiments

pratiksha:
Code to run accuracy experiments for string histograms and mixed-column heatmaps, and save to file.


web/src/main/java/org/hillview/dataStructures/IntervalDecomposition.java
```diff
@@ -155,8 +157,8 @@ public int getQuantizationIntervalCount() {
     @SuppressWarnings("SameParameterValue")
     List<Pair<Integer, Integer>> bucketDecomposition(int bucketIdx, boolean cdf) {
         Pair<Integer, Integer> range = this.bucketRange(bucketIdx, cdf);
-        return IntervalDecomposition.dyadicDecomposition(
-            Converters.checkNull(range.first), Converters.checkNull(range.second));
+        return IntervalDecomposition.kadicDecomposition(
```
mbudiu-vmw: Always?
No policy to decide?
We should add this to the JSON file as well. Would you like to do it in this PR?

web/src/main/java/org/hillview/dataStructures/PrivateHeatmap.java
```diff
@@ -46,8 +61,12 @@ private void noiseForBucket(
         }
     }
 
-    private static boolean notConfident(long value, int noise) {
-        return value < noise;
+    public void noiseForRange(int left, int right, int top, int bot,
+                               double scale, double baseVariance, /*out*/Noise result) {
+        List<Pair<Integer, Integer>> xIntervals = dx.kadicDecomposition(left, right, IntervalDecomposition.BRANCHING_FACTOR);
+        List<Pair<Integer, Integer>> yIntervals = dy.kadicDecomposition(top, bot, IntervalDecomposition.BRANCHING_FACTOR);
```
mbudiu-vmw: Yes, let's not make BRANCHING_FACTOR a constant, it should be a field.

web/src/test/java/org/hillview/HistogramAccuracyTest.java
```diff
@@ -160,14 +175,71 @@ private double computeAccuracy(PrivateHistogram ph, int totalLeaves) {
...
+     * Compute the absolute and L2 error vs. ground truth for an instantiation of a private heat map
+     * averaged over all possible range queries (every rectangle).
+     *
+     * @param ph The heat map whose accuracy we want to compute.
+     * @param totalXLeaves The "global" number of leaves (in the x-dimension) in case this heat map is computed only on a zoomed-in range.
+     *                    This is needed to correctly compute the amount of noise to add to each leaf.
+     * @param totalYLeaves same as totalXLeaves but for y-axis
+     * @return The average per-query absolute error.
+     */
+    private Double computeAccuracy(PrivateHeatmap ph, int totalXLeaves, int totalYLeaves) {
```
mbudiu-vmw: I expect that you will want to use this kind of code in the benchmarking program, so it should probably become a method - would it work in the PrivateHeatmap?

We should also add the code that is used for measurements to the DPBenchmarks class - perhaps with more command-line arguments to choose one of the benchmarks to run. Ideally we should be able to run all the measurements with just one program.

web/src/test/java/org/hillview/HistogramAccuracyTest.java
```diff
@@ -203,20 +320,81 @@ public void computeAccuracyTest() {
...
+
+            // for JSON parsing convenience
+            ArrayList<Double> resArr = new ArrayList();
+            resArr.add(res.first); // noise
+            resArr.add(res.second); // stdev
+
+            results.put(col, resArr);
         }
+
+        FileWriter writer = new FileWriter(histogram_results_filename);
```
mbudiu-vmw: I general it's not a good idea to have tests leave stuff like this on the filesystem.
You should clean-up at the end. You can use File.deleteOnExit, for example.
If you want persistent results then we should move this code to DPBenchmarks.

mbudiu-vmw:
I was also expecting some changes to DPBenchmarks to run string measurements as well.

web/src/main/java/org/hillview/dataStructures/PrivateHeatmap.java
pratiksha: Not sure what you mean by this, is there something you want to change here? I don't want it to be a variable that gets passed around; I think it should be a constant on the server.

web/src/main/java/org/hillview/dataStructures/IntervalDecomposition.java
pratiksha: It can be a policy parameter, but I don't think adding this to the schema needs to be a priority right now; I can open an issue for it and we can address this later.

web/src/main/java/org/hillview/dataStructures/PrivateHeatmap.java
pratiksha: See comment above.

web/src/test/java/org/hillview/HistogramAccuracyTest.java
pratiksha: Hmm, I don't think this should be functionality of the histogram/heatmap as it's really benchmarking code, e.g. we'll never expose this benchmark information to a user.

I'll move this to a benchmark class.

pratiksha:
@mbudiu-vmw PTAL

PR #594, opened 2020-01-18
Add column index to PRF input

pratiksha:
As title.


platform/src/main/java/org/hillview/table/QuantizationSchema.java
```diff
@@ -32,6 +32,9 @@
     // We use a LinkedHashMap for deterministic serialization
     private LinkedHashMap<String, ColumnQuantization> quantization;
 
+    @Nullable
+    private LinkedHashMap<String, Integer> columnIndexes;
```
mbudiu-vmw: If you mutate the schema you should make the methods `@synchronized` - otherwise there's chance of concurrent clients doing it concurrently.

web/src/main/java/org/hillview/dataStructures/PrivateHistogram.java
```diff
@@ -21,11 +21,14 @@
     // TODO(pratiksha): compute the missing value confidence
     private int missingConfidence;  // Confidence for the missing value
     private final double epsilon;
+    private final int columnIndex;
```
mbudiu-vmw: The histogram is serialized to the UI, we should not store in it fields that are not needed by the UI.
We'll have to dismantle this class and build it more similar to the Heatmap instead.

mbudiu-vmw:
I fixed the conflict.

PR #596, opened 2020-01-30
Empirical confidence intervals

pratiksha:
Implements confidence intervals by sampling the CDF and computing the empirical percentiles.


pratiksha:
Let's maybe discuss this before merging; it's probably also possible to just use a Chernoff bound directly to get this interval.

PR #597, opened 2020-02-13
Support for postprocessing; cleanup private histograms

mbudiu-vmw:


mbudiu-vmw:
@pratiksha : this introduces a framework to do postprocessing on sketches in a more systematic way, and rewrites histograms and heatmaps to use it.

PR #598, opened 2020-03-05
fixed tomcat start on Windows

vbondzio:
On Windows (at least without WLS), its catalina.bat instead of .sh. Also set CATALINA_HOME since tomcat guessing it didn't work for some reason.


vbondzio:
adding myself to the vmware org first ...

PR #606, opened 2020-05-27
Track provenance with each UI page

### LINKED ISSUE #605: Title wrong after zooming in heat map
When creating a heatmap, and then zooming, the title of the new heatmap is object/object, instead of the actual column names. 
The issue was first seen on a weather data. File was loaded as .csv, schema was inferred, column names came from first row. Types of columns are Double/Integer.

mbudiu-vmw:
Also fixes #605 
Improves date parsing.


mbudiu-vmw:
@uwieder : please try this version for improved date parsing

PR #621, opened 2020-06-24
Generalized DataUpload to also support grok logs

mbudiu-vmw:


mbudiu-vmw:
@shehbazj: this is a tool which splits a large file into multiple small files; it can output either csv or orc formats currently.

PR #633, opened 2020-07-02
Adding test case and remove redundant code at columns loader

daniarherikurniawan:
Now the lazy loader and non-lazy loader will point to loadColumns()

### DIFF
--- a/platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
+++ b/platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
@@ -18,39 +18,39 @@
 package org.hillview.storage;
 
 import org.hillview.table.api.*;
+import org.hillview.table.columns.BaseListColumn;
+import org.hillview.table.ColumnDescription;
+import org.hillview.table.Schema;
+import org.hillview.table.Table;
+import org.hillview.utils.Converters;
+import org.hillview.utils.HillviewLogger;
+import org.hillview.utils.Linq;
+
+import org.apache.cassandra.utils.FBUtilities;
 import org.apache.cassandra.config.CFMetaData;
 import org.apache.cassandra.config.ColumnDefinition;
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.cql3.ColumnIdentifier;
 import org.apache.cassandra.db.rows.Cell;
 import org.apache.cassandra.db.rows.ColumnData;
-import org.apache.cassandra.db.rows.RangeTombstoneMarker;
 import org.apache.cassandra.db.rows.Row;
 import org.apache.cassandra.db.rows.Unfiltered;
 import org.apache.cassandra.db.rows.UnfilteredRowIterator;
-import org.apache.cassandra.io.sstable.Descriptor;
-import org.apache.cassandra.io.sstable.ISSTableScanner;
-import org.apache.cassandra.io.sstable.format.SSTableReader;
-import org.hillview.table.ColumnDescription;
-import org.hillview.table.Schema;
-import org.hillview.table.Table;
-import org.hillview.table.membership.FullMembershipSet;
-import org.hillview.utils.Converters;
-import org.hillview.utils.HillviewLogger;
-
-import javax.annotation.Nullable;
-import java.util.stream.Stream;
-import java.util.*;
-import java.util.stream.StreamSupport;
-
 import org.apache.cassandra.db.PartitionColumns;
 import org.apache.cassandra.db.SerializationHeader;
 import org.apache.cassandra.db.marshal.AbstractType;
 import org.apache.cassandra.db.marshal.UTF8Type;
 import org.apache.cassandra.dht.*;
+import org.apache.cassandra.io.sstable.Descriptor;
+import org.apache.cassandra.io.sstable.ISSTableScanner;
+import org.apache.cassandra.io.sstable.format.SSTableReader;
 import org.apache.cassandra.io.sstable.metadata.MetadataComponent;
 import org.apache.cassandra.io.sstable.metadata.MetadataType;
-import org.apache.cassandra.utils.FBUtilities;
+
+import javax.annotation.Nullable;
+import java.util.stream.Stream;
+import java.util.*;
+import java.util.stream.StreamSupport;
 
 /**
  * Knows how to read a local Cassandra's SSTable file.
@@ -63,16 +63,15 @@ public class CassandraSSTableLoader extends TextFileLoader{
     IAppendableColumn[] columns;
 
     public boolean lazyLoading;
-    private int currentColumn;
     private long rowCount;
     private final CFMetaData metadata;
     private final SSTableReader ssTableReader;
 
     static{
-        /* Initializing DatabaseDescriptor that contains Cassandra's general configuration */
+        // Initializing DatabaseDescriptor that contains Cassandra's general configuration
         DatabaseDescriptor.clientInitialization(false);
     }
-    
+
     public CassandraSSTableLoader(String ssTablePath, boolean lazyLoading) {
         super(ssTablePath);
         this.ssTablePath = ssTablePath;
@@ -82,10 +81,11 @@ public class CassandraSSTableLoader extends TextFileLoader{
             Descriptor descriptor = Descriptor.fromFilename(this.ssTablePath);
             this.metadata = this.getSSTableMetadata(descriptor);
             this.ssTableReader = SSTableReader.openNoValidation(descriptor, this.metadata);
-            /* Get the schema of the current SSTable */
+            // Get the schema of the current SSTable
             this.actualSchema = this.getSchema(this.metadata);
         } catch (Exception e) {
-            HillviewLogger.instance.error("Failed initializing Metadata and SStable partitions based on: " + this.ssTablePath);
+            HillviewLogger.instance.error("Failed initializing Metadata and SStable partitions based on: " 
+                + this.ssTablePath);
             throw new RuntimeException(e);
         }
     }
@@ -188,54 +188,55 @@ public class CassandraSSTableLoader extends TextFileLoader{
         private final Schema actualSchema;
         private int sizeToLoad;
 
-        SSTableColumnLoader(SSTableReader ssTableReader, Schema actualSchema){
+        SSTableColumnLoader(SSTableReader ssTableReader, Schema actualSchema) {
             this.ssTableReader = ssTableReader;
             this.actualSchema = actualSchema;
         }
 
-        /** Marking the corresponding column index that are included in List<String> names */
-        private boolean[] getColumnMarker(List<String> names){
+        // Marking the corresponding column index that are included in List<String> names
+        private boolean[] getColumnMarker(List<String> names) {
             this.sizeToLoad = 0;
             List<String> colNames = this.actualSchema.getColumnNames();
             boolean[] result = new boolean[colNames.size()];
             int i = 0;
             for (String cn : colNames) {
-                if(names.contains(cn)){
+                if(names.contains(cn)) {
                     result[i] = true;
-                    /* Incrementing the number of column that will be loaded  */
+                    // Incrementing the number of column that will be loaded
                     this.sizeToLoad++;
-                }else{
+                } else {
                     result[i] = false;
                 }
                 i++;
             }
             return result;
         }
-        
+
         @Override
         public List<IColumn> loadColumns(List<String> names) {
-            IAppendableColumn[] columns = this.actualSchema.createAppendableColumns();
-            boolean[] columnToLoad = getColumnMarker(names);
+            boolean[] columnToLoad = this.getColumnMarker(names);
             ISSTableScanner currentScanner = this.ssTableReader.getScanner();
-            Spliterator<UnfilteredRowIterator> splititer = Spliterators.spliteratorUnknownSize(currentScanner, Spliterator.IMMUTABLE);
+            Spliterator<UnfilteredRowIterator> splititer = Spliterators.spliteratorUnknownSize(currentScanner, 
+                Spliterator.IMMUTABLE);
             Stream<UnfilteredRowIterator> partitions = StreamSupport.stream(splititer, false);
 
-            return CassandraSSTableLoader.this.loadColumns(columns, columnToLoad, partitions, this.sizeToLoad );
+            return CassandraSSTableLoader.this.loadColumns(columnToLoad, partitions, this.sizeToLoad);
         }
     }
 
     public int getNumRows() {
         ISSTableScanner currentScanner = this.ssTableReader.getScanner();
-        Spliterator<UnfilteredRowIterator> splititer = Spliterators.spliteratorUnknownSize(currentScanner, Spliterator.IMMUTABLE);
+        Spliterator<UnfilteredRowIterator> splititer = Spliterators.spliteratorUnknownSize(currentScanner, 
+            Spliterator.IMMUTABLE);
         Stream<UnfilteredRowIterator> partitions = StreamSupport.stream(splititer, false);
-        /* Iterating each item inside the table */
+        // Iterating each item inside the table
         partitions.forEach(partition -> {
             Unfiltered unfiltered;
-            /* Serializing each partition */
-            while (partition.hasNext()){
+            // Serializing each partition
+            while (partition.hasNext()) {
                 unfiltered = partition.next();
-                /* Hillview ignores Tombstone Markers, just count the Row */
-                if (unfiltered instanceof Row){
+                // Hillview ignores Tombstone Markers, just count the Row
+                if (unfiltered instanceof Row) {
                     this.rowCount++;
                 }
             }
@@ -243,9 +244,23 @@ public class CassandraSSTableLoader extends TextFileLoader{
         return Converters.toInt(rowCount);
     }
 
-    public ITable load(){
+    /** Instead of checking the columnn' name to find which one to load,
+     * this method uses boolean marker stored at columnToLoad to recognize the needed columns */
+    private List<IAppendableColumn> createColumns(boolean[] columnToLoad, int sizeToLoad) {
+        List<ColumnDescription> cols = this.actualSchema.getColumnDescriptions();
+        List<IAppendableColumn> result = new ArrayList<IAppendableColumn>(sizeToLoad);
+        int i = 0;
+        for (ColumnDescription cd : cols) {
+            if (columnToLoad[i])
+                result.add(BaseListColumn.create(cd));
+            i++;
+        }
+        return result;
+    }
+
+    /** This is the general column loader that can be triggered once the initialization is complete */
+    public ITable load() {
         try {
-            this.columns = this.actualSchema.createAppendableColumns();
             if (this.lazyLoading) {
                 int size = this.getNumRows();
                 PartitionColumns columnDefinitions = metadata.partitionColumns();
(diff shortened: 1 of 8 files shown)


platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -215,67 +214,13 @@ public Schema getSchema(CFMetaData metadata) {
         
         @Override
         public List<IColumn> loadColumns(List<String> names) {
-            boolean[] columnToLoad = getColumnMarker(names);
             IAppendableColumn[] columns = this.actualSchema.createAppendableColumns();
+            boolean[] columnToLoad = getColumnMarker(names);
```
mbudiu-vmw, line 218: In general we prefer `this.` for any local field, including methods. This is actually mandatory in a language like typeScript, and I think it's very useful.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -215,67 +214,13 @@ public Schema getSchema(CFMetaData metadata) {
...
-            for (int idx = 0; idx < columns.length; idx++) {
-                appendableColumn = columns[idx];
-                column = appendableColumn.seal();
-                if (membershipSet == null)
-                    membershipSet = new FullMembershipSet(column.sizeInRows());
-                if(columnToLoad[idx])
-                    arrColumns.add(column);
-            }
-            return arrColumns;
+            return CassandraSSTableLoader.this.loadColumns(columns, columnToLoad, partitions, this.sizeToLoad );
```
mbudiu-vmw, line 223: no space before last parens

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -312,21 +258,25 @@ public ITable load(){
                 IColumnLoader loader = new CassandraSSTableLoader.SSTableColumnLoader(this.ssTableReader, this.actualSchema);
                 return Table.createLazyTable(cds, size, this.filename, loader);
             } else {
-                return nonLazyLoading();
+                int sizeToLoad = this.columns.length;
+                /** columns loader will load all column, so all item of columnToLoad need to be TRUE */
```
mbudiu-vmw, line 262: do not use JavaDoc comments in the code; they are reserved for declarations only.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -312,21 +258,25 @@ public ITable load(){
...
+                boolean[] columnToLoad = new boolean[sizeToLoad];
+                Arrays.fill(columnToLoad, Boolean.TRUE);
+
+                ISSTableScanner currentScanner = this.ssTableReader.getScanner();
+                Spliterator<UnfilteredRowIterator> splititer = Spliterators.spliteratorUnknownSize(currentScanner, Spliterator.IMMUTABLE);
+                Stream<UnfilteredRowIterator> partitions = StreamSupport.stream(splititer, false);
+
+                List<IColumn> arrColumns = loadColumns(this.columns ,columnToLoad, partitions, sizeToLoad);
```
mbudiu-vmw, line 270: no space before comma

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -345,42 +295,42 @@ public ITable nonLazyLoading() {
...
                     /* Extracting the value for each column from a single row  */
                     for (ColumnData cd : row) {
-                        if (cd.column().isSimple()) {
-                            cell = (Cell) cd;
-                            cellType = cell.column().cellValueType();
-                            value = cellType.getSerializer().deserialize(cell.value());
-                        } else {
-                            throw new RuntimeException("[Hillview can't convert complex data]");
+                        /* Filter the column to only load the one marked true by columnToLoad[] */
+                        if (columnToLoad[i]){
```
mbudiu-vmw, line 299: space before {

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -345,42 +295,42 @@ public ITable nonLazyLoading() {
                     i = 0;
                     /* Extracting the value for each column from a single row  */
                     for (ColumnData cd : row) {
-                        if (cd.column().isSimple()) {
-                            cell = (Cell) cd;
-                            cellType = cell.column().cellValueType();
-                            value = cellType.getSerializer().deserialize(cell.value());
-                        } else {
-                            throw new RuntimeException("[Hillview can't convert complex data]");
+                        /* Filter the column to only load the one marked true by columnToLoad[] */
```
mbudiu-vmw, line 298: this comment is not very useful; instead you can document the function using JavaDoc and add a description of all arguments.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -345,42 +295,42 @@ public ITable nonLazyLoading() {
...
+                        if (columnToLoad[i]){
+                            if (cd.column().isSimple()){
```
mbudiu-vmw, line 300: space

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -345,42 +295,42 @@ public ITable nonLazyLoading() {
...
+                                cell = (Cell) cd;
+                                cellType = cell.column().cellValueType();
+                                value = cellType.getSerializer().deserialize(cell.value());
+                            }else{
```
mbudiu-vmw, line 304: spaces around else

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -345,42 +295,42 @@ public ITable nonLazyLoading() {
...
+                                /* Hillview won't process the complex data */
```
mbudiu-vmw, line 305: This is only a temporary solution, and probably you should throw this exception when you create the schema.
We should probably replace such columns with String values that use value.toString() as contents.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -312,21 +258,25 @@ public ITable load(){
...
         } catch (Exception ex) {
             throw new RuntimeException(ex);
         }
     }
     
-    public ITable nonLazyLoading() {
-        assert this.actualSchema != null;
-        this.columns = this.actualSchema.createAppendableColumns();
-
+    private List<IColumn> loadColumns( IAppendableColumn[] columns, boolean[] columnToLoad, Stream<UnfilteredRowIterator> partitions, int sizeToLoad){
```
mbudiu-vmw, line 278: no space after (, but add a space before {

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -345,42 +295,42 @@ public ITable nonLazyLoading() {
...
-
                     /* Append each line to Hillview column */
                     for (this.currentColumn = 0; this.currentColumn < line.length; this.currentColumn++) {
-                        this.columns[this.currentColumn].append(line[this.currentColumn]);
+                        columns[this.currentColumn].append(line[this.currentColumn]);
                     }
-                    this.currentRow++;
-
-                }else if (unfiltered instanceof RangeTombstoneMarker){
+                } else if (unfiltered instanceof RangeTombstoneMarker) {
```
mbudiu-vmw, line 316: What are these? Is there a reason to process them? Do they indicate a deleted row?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -345,42 +295,42 @@ public ITable nonLazyLoading() {
...
 
-        IColumn[] arrColumns = new IColumn[this.columns.length];
+        List<IColumn> arrColumns = new ArrayList<IColumn>(sizeToLoad);
         IMembershipSet membershipSet = null;
         IAppendableColumn appendableColumn;
         IColumn column;
-        for (int idx = 0; idx < this.columns.length; idx++) {
-            appendableColumn = this.columns[idx];
+        for (int idx = 0; idx < columns.length; idx++) {
+            appendableColumn = columns[idx];
```
mbudiu-vmw, line 327: where is the membershipSet used?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -345,42 +295,42 @@ public ITable nonLazyLoading() {
...
         });
 
-        IColumn[] arrColumns = new IColumn[this.columns.length];
+        List<IColumn> arrColumns = new ArrayList<IColumn>(sizeToLoad);
         IMembershipSet membershipSet = null;
         IAppendableColumn appendableColumn;
         IColumn column;
-        for (int idx = 0; idx < this.columns.length; idx++) {
-            appendableColumn = this.columns[idx];
+        for (int idx = 0; idx < columns.length; idx++) {
```
mbudiu-vmw, line 326: I don't understand what happens here. You should only load the columns you were asked, why are you copying some of them? What's in the other ones?

platform/src/test/java/org/hillview/test/storage/SSTableTest.java
```diff
@@ -53,11 +53,21 @@ public void testSSTableComplimentaryFiles() {
     @Test
 
     public void testReadingSSTable() {
-        CassandraSSTableLoader ssTableLoader = new CassandraSSTableLoader(this.ssTablePath, false);
+        boolean lazyLoading = false;
+        CassandraSSTableLoader ssTableLoader = new CassandraSSTableLoader(this.ssTablePath, lazyLoading);
         try {
             ITable table = ssTableLoader.load();
+            String name = table.getLoadedColumn("name").getString(6);
```
mbudiu-vmw, line 60: make 6 a final constant?

platform/src/test/java/org/hillview/test/storage/SSTableTest.java
```diff
@@ -53,11 +53,21 @@ public void testSSTableComplimentaryFiles() {
...
+            String address = table.getLoadedColumn("address").getString(6);
+            int salary = table.getLoadedColumn("salary").getInt(6);
+            String phone = table.getLoadedColumn("phone").getString(6);
+
             Assert.assertNotNull(table);
             Assert.assertEquals("Table[4x15]", table.toString());
+            Assert.assertEquals("azam", name);
+            Assert.assertEquals("Chennai", address);
+            Assert.assertEquals(45000, salary);
+            Assert.assertEquals("9848022330", phone);
```
mbudiu-vmw, line 70: If you checked-in this file there is no reason this test should fail, so perhaps you can remove all these catch

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: Noted! I made some changes

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: Good catch! I am checking others

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: Will do!

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: Got it!

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
mbudiu-vmw: no need to reply to every comment, just mark them as resolved

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: Fixed it and others as well

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: Yes, it's a marker that the item is deleted

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: This is my mistake, membershipSet is not being used here. The other class uses membership set, and I forgot to delete it when I use it here

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: I will remove this. I was using  this.actualSchema.createAppendableColumns(); to create the columns, and it returns IAppendableColumn[] instead of List<IColumn> . The copying is to convert the final result into List. I will update the code to avoid copying

platform/src/test/java/org/hillview/test/storage/SSTableTest.java
daniarherikurniawan: Right, I will remove them

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -309,78 +270,77 @@ public ITable load(){
...
-        Stream<UnfilteredRowIterator> partitions = StreamSupport.stream(splititer, false);
+    /** Lazy and non-lazy loader will call this function to load specific column marked by columnToLoad */
+    private List<IColumn> loadColumns(boolean[] columnToLoad, Stream<UnfilteredRowIterator> partitions, int size) {
+        List<IAppendableColumn> columns = createColumns(columnToLoad, size);
+        // Iterating each item inside the table
         partitions.forEach(partition -> {
             int i;
             Object value;
-            Object[] line;
+            List<Object> line;
```
mbudiu-vmw, line 302: can some of these declarations be pushed inside closer to their use?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -309,78 +270,77 @@ public ITable load(){
...
             Row row;
             Cell cell;
             AbstractType<?> cellType;
             Unfiltered unfiltered;
 
-            /* Serializing each partition */
-            while (partition.hasNext()){
+            // Serializing each partition
```
mbudiu-vmw, line 308: this kind of comment does not add much, it is rather obvious what happens on the next line

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -309,78 +270,77 @@ public ITable load(){
...
+            while (partition.hasNext()) {
                 unfiltered = partition.next();
-                if (unfiltered instanceof Row){
+                if (unfiltered instanceof Row) {
                     row = (Row) unfiltered;
-                    line = new Object[row.columnCount()];
+                    line = new ArrayList<Object>(row.columnCount());
```
mbudiu-vmw, line 313: why do you need this array?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -309,78 +270,77 @@ public ITable load(){
...
+                        if (columnToLoad[i]) {
+                            if (cd.column().isSimple()) {
+                                cell = (Cell) cd;
+                                cellType = cell.column().cellValueType();
+                                value = cellType.getSerializer().deserialize(cell.value());
+                            } else {
+                                // Hillview won't process the complex data
+                                throw new RuntimeException("[Hillview can't convert complex data]");
+                            }
+                            line.add(value);
```
mbudiu-vmw, line 327: why not append directly to the column instead of adding it to the line?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -309,78 +270,77 @@ public ITable load(){
...
-            if (membershipSet == null)
-                membershipSet = new FullMembershipSet(column.sizeInRows());
-            arrColumns[idx] = column;
-            assert arrColumns[idx] != null;
-        }
-
-        return new Table(arrColumns, this.ssTablePath, null);
+        for (IAppendableColumn c : columns)
+            c.seal();
+        return Linq.map(columns, e -> e);
```
mbudiu-vmw, line 344: this is really useless; just return columns.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
mbudiu-vmw: In general, you should avoid allocating objects in inner loops if not absolutely necessary; this will generate a lot of useless garbage.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -183,122 +183,83 @@ public Schema getSchema(CFMetaData metadata) {
...
             Unfiltered unfiltered;
-            /* Serializing each partition */
-            while (partition.hasNext()){
+            // Serializing each partition
+            while (partition.hasNext()) {
                 unfiltered = partition.next();
-                /* Hillview ignores Tombstone Markers, just count the Row */
-                if (unfiltered instanceof Row){
+                // Hillview ignores Tombstone Markers, just count the Row
+                if (unfiltered instanceof Row) {
```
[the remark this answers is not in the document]
daniarherikurniawan: I will use partitions.count(). However, this value might not be accurate since a partition not always contains a valid row, it could be a tombstone marker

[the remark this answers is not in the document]
mbudiu-vmw: doesn't the partition maintain a tombstone count then?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: I removed the "for" loop. However, the "Linq.map" is still needed to convert List<IAppendableColumn> to List<IColumn>

mbudiu-vmw: The right solution is probably to change the API of IColumnLoader to return List<T extends IColumn>. Maybe you can try to do this.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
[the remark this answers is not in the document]
daniarherikurniawan: There is no high-level API, but I can trace how Cassandra gets the table stats. The tracing result will include an estimated number of rows and tombstone counts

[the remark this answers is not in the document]
mbudiu-vmw: In principle I think the row count to be exact, I don't know what would happen if we return an approximation.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: Okay! This will trigger some changes in other classes, such as JdbcDatabase and ParquetFileLoader

PR #635, opened 2020-07-07
Add Cassandra DB connector

daniarherikurniawan:
1. One of the unit test depends on my local Cassandra setup, thus I comment it out. I will add a Cassandra installer later.
2. CassandraConnectionInfo provides the needed parameters to establish connection with Cassandra node.
3. CassandraDatabase is able to get the available keyspaces (+ tables), get the keyrange (+ replica of each range), and get the sstable path of any available table.
4. There are two type of connection:
- NodeProbe connection, for running server-side query (getting the replica of each key-range)
- Cluster connection, for running client-side query  (getting keyspace’s metadata, numRows, etc)
5. Inside CassandraDatabase, there are two internal classes: CassTable and TokenRange. I put them inside CassandraDatabase because they are too small. CassTable holds the value of available keyspace and its tables.  And, TokenRange holds the range of the token and the endpoints that store each range. 
6. There is bit of parsing inside TokenRange constructor, due to the output from describering <keyspace> can't be converted to yaml or json. The rawEntry looks like this:

TokenRange(start_token:-2927249996180110524, end_token:-2814694817804652701, endpoints:[155.98.36.95, 155.98.36.91], rpc_endpoints:[155.98.36.95, 155.98.36.91], endpoint_details:[EndpointDetails(host:155.98.36.95, datacenter:datacenter1, rack:rack1), EndpointDetails(host:155.98.36.91, datacenter:datacenter1, rack:rack1)])
 
I believe this parsing won't cause any overhead since there are only 256 (default) token-range. The recommended size for real deployment is only 16 token range.


platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -0,0 +1,241 @@
...
+            p_stdin.newLine();
+            p_stdin.flush();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+
+        // scan the output of executed commands
+        Scanner s = new Scanner(p.getInputStream());
+        while (s.hasNext()) {
+            sstablePath = s.next();
```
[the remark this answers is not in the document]
daniarherikurniawan: Yes. The output are two lines. The first line is not important

platform/src/test/java/org/hillview/test/storage/SSTableTest.java
```diff
@@ -114,4 +129,41 @@ public void testLazyLoading() {
...
+            // Connecting to Cassandra node and get some data
+                CassandraDatabase cassDB = new CassandraDatabase(this.info);
+                cassDB.loadKeyspacesAndTables();
+                cassDB.printCassTables();
+                cassDB.loadTablePartition("cassdb");
+                cassDB.printTablePartition();
+                long count = cassDB.getRowCount("cassDB", "flights");
+                System.out.println("Count: " + count + " rows");
+
+                String ssTablePath = cassDB.getSSTablePath("cassdb", "flights");
```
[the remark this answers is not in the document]
daniarherikurniawan: Since we put the database name and table name in the ConnectionInfo, I believe the client will provide us the correct information regarding those two fields. Do we still need a method to discover the available databases and tables in the cluster? Perhaps we need to display this databases to the client so that they can choose. 

The table partition will be used by SSTableReader to load the local key-range only. These three methods are independent of each other: loadKeyspacesAndTables(), loadTablePartition(), and getSSTablePath()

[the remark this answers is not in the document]
mbudiu-vmw: I thought about this, and having a browser could be useful, but for now it's not our most urgent feature, so let's assume that the users can use other Cassandra tools to browse the database. So let's keep the implementation as simple as possible.

bin/install-cassandra.sh
```diff
@@ -0,0 +1,46 @@
...
+mkdir $ROOT_DIR/cassandra/logs
+
+# Install Java 8 [Run this if you don't have it]
+# sudo apt-get update
+# printf "Y" | sudo apt-get install openjdk-8-jdk
+
+# Start Cassandra
+$ROOT_DIR/cassandra/bin/cassandra
+
+# Generate 100 flights data (adding unique primary key) from the existing one
```
mbudiu-vmw, line 21: Can't we use autoincrement instead of generating a primary key? For 100 rows it should not take too long.

bin/install-cassandra.sh
```diff
@@ -0,0 +1,46 @@
+SAVEDIR=$PWD
```
mbudiu-vmw, line 1: Add a `#!/bin/bash` and also a comment explaining what tihs file does.

bin/install-cassandra.sh
```diff
@@ -0,0 +1,46 @@
...
+# Change this ROOR_DIR to your preferred instalation directory, and make sure to update cassandraRootDir at SSTableTest.java
+ROOT_DIR="/tmp/"
```
mbudiu-vmw, line 3: Call this CASSANDRA_INSTALLATION_DIR

bin/install-cassandra.sh
```diff
@@ -0,0 +1,46 @@
...
+
+[ -e $ROOT_DIR/apache-cassandra-3.11.6-bin.tar.gz ] && rm $ROOT_DIR/apache-cassandra-3.11.6-bin.tar.gz
```
mbudiu-vmw, line 5: why delete it if it exists? Perhaps you should do just the opposite, only download it if it does not exist.
Maybe you can make 3.11.6 a CASSANDRA_VERSION variable.

bin/install-cassandra.sh
```diff
@@ -0,0 +1,46 @@
...
+[ -e $ROOT_DIR/cassandra ] && rm -rf $ROOT_DIR/cassandra
+
+# Download Cassandra to $ROOT_DIR
+wget https://downloads.apache.org/cassandra/3.11.6/apache-cassandra-3.11.6-bin.tar.gz -O $ROOT_DIR/apache-cassandra-3.11.6-bin.tar.gz
+tar xzvf $ROOT_DIR/apache-cassandra-3.11.6-bin.tar.gz -C $ROOT_DIR
+mv $ROOT_DIR/apache-cassandra-3.11.6 $ROOT_DIR/cassandra
+mkdir $ROOT_DIR/cassandra/logs
+
+# Install Java 8 [Run this if you don't have it]
```
mbudiu-vmw, line 14: This is required for Hillview anyway; you should probably delete this.

bin/install-cassandra.sh
```diff
@@ -0,0 +1,46 @@
...
+python3 ../data/sstable/generate-data.py
+
+echo "Waiting for 10s to let Cassandra finish the initialization"
+sleep 10
+
+# Check Cassandra status
+$ROOT_DIR/cassandra/bin/nodetool status
+
+# Load data to Cassandra
+if [[ $(wc -l < ../data/sstable/cassdb.cql) -le 93 ]]; then
```
mbudiu-vmw, line 31: why do you have to modify this script?
Can't you just execute a different script to load the other dataset?
Move this stuff into the demo-data-cleaner as well.

bin/install-cassandra.sh
```diff
@@ -0,0 +1,46 @@
...
+    # this will append a relative path to load the csv data
+    echo " FROM '$SAVEDIR/../data/sstable/flights_data.csv' WITH DELIMITER=',' AND HEADER=TRUE; exit;" >> ../data/sstable/cassdb.cql
+fi
+
+$ROOT_DIR/cassandra/bin/cqlsh --file ../data/sstable/cassdb.cql
+
+# Force Compaction to put the data into sstable
+$ROOT_DIR/cassandra/bin/nodetool flush
+$ROOT_DIR/cassandra/bin/nodetool compact cassdb
```
mbudiu-vmw, line 40: these should be in the data script as well probably

data/sstable/cassdb.cql
```diff
@@ -0,0 +1,94 @@
...
+   salary, phone) VALUES(now(),'', '', 57800, '9848022338');
+
+INSERT INTO users (id, name, address,
+   salary, phone) VALUES(now(),'', '', 29000, '');
+
+INSERT INTO users (id, name, address,
+   salary, phone) VALUES(now(),'', '',0, '');
+
+
+CREATE TABLE flights(
```
mbudiu-vmw, line 75: make this a separate script to load data.

data/sstable/generate-data.py
```diff
@@ -0,0 +1,15 @@
+import os.path
```
mbudiu-vmw, line 1: hopefully this won't be needed.

platform/src/main/java/org/hillview/storage/CassandraConnectionInfo.java
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.hillview.storage;
+
+public class CassandraConnectionInfo extends JdbcConnectionInformation {
+    static final long serialVersionUID = 1;
+
+    // Port for establising probe connection
```
mbudiu-vmw, line 23: typo in comment

platform/src/main/java/org/hillview/storage/CassandraConnectionInfo.java
```diff
@@ -0,0 +1,44 @@
...
+    public int jmxPort;
+    public String cassandraRootDir;
```
mbudiu-vmw, line 25: Please comment this field too. Here you should use JavaDoc `/** */`

platform/src/main/java/org/hillview/storage/CassandraConnectionInfo.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public CassandraConnectionInfo (String cassandraRootDir, String host, int jmxPort,
+        int nativePort, String keyspace, String table, String user, String password){
+        this.cassandraRootDir = cassandraRootDir;
```
mbudiu-vmw, line 29: Do you want to add a constructor for JdbcConnectionInformation too? It wasn't needed so far, because that class is now read using serialization from json.

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -0,0 +1,264 @@
...
+    public CassandraDatabase(CassandraConnectionInfo info) {
+        this.info = info;
+        try {
+            this.nodeProbeFactory = new NodeProbeFactory();
+            this.connectLocalProbe();
+            this.connectCassCluster();
+            this.setStoredTableInfo();
+            this.setTokenRanges();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
mbudiu-vmw, line 88: How about you only log an error here and rethrow?

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /** Connection for running server-side query. */
+    private void connectLocalProbe() {
+        try {
+            if (Utilities.isNullOrEmpty(info.user)) {
+                this.probe = nodeProbeFactory.create(info.host, info.jmxPort);
+            } else
+                this.probe = nodeProbeFactory.create(info.host, info.jmxPort,
+                    info.user, info.password);
+        } catch (IOException | SecurityException e) {
```
mbudiu-vmw, line 121: instead of catching the exceptions here add them to the method signature and let the caller catch them.

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -0,0 +1,264 @@
...
+        }
+    }
+
+    public static class TokenRange {
+        public String start_token;
+        public String end_token;
+        public List<String> endpoints = new ArrayList<>();
+
+        public TokenRange(String rawEntry) {
+            try{
```
mbudiu-vmw, line 135: space

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -0,0 +1,264 @@
...
+     */
+    private void setTokenRanges() {
+        String keyspace = this.info.database;
+        try {
+            TokenRange tr;
+            for (String tokenRangeString : this.probe.describeRing(keyspace)) {
+                tr = new TokenRange(tokenRangeString);
+                this.tokenRanges.add(tr);
+            }
+        } catch (IOException e) {
```
mbudiu-vmw, line 169: don't catch this one

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -0,0 +1,264 @@
...
+        public String keyspace;
+        public List<String> tables;
+
+        public CassTable(String keyspace, List<String> tables) {
+            this.keyspace = keyspace;
+            this.tables = tables;
+        }
+
+        public String toString(){
+            String result = this.keyspace;
```
mbudiu-vmw, line 184: space. Also, don't use String concatenation in a loop, use a StringBuilder instead.

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -0,0 +1,264 @@
...
+     * Using sstableutil command is better than doing manually sstable search. Although we know
+     * the sstable parent dir from cassandra.yaml, we need to rule out the outdated sstables.
+     * And sstableutil is built specifically to identify which version are the most updated.
+      */
+    public String getSSTablePath() {
+        String sstablePath = EMPTY;
+        boolean isWindows = System.getProperty("os.name")
+            .toLowerCase().startsWith("windows");
+        ProcessBuilder builder = new ProcessBuilder();
+        if (isWindows) {
```
mbudiu-vmw, line 222: how about you don't use a shell, just call that command, and instead implement the grep in java on the standard output? Then you don't care whether this is windows or linux, it's the same command.

mbudiu-vmw: Paths to create the path, then you don't care whether it's "/" or "\".

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    public Long getRowCount(String keyspace, String table) {
+        ResultSet result = this.session.execute("SELECT count(*) FROM " + keyspace + "." + table);
+        Row r = result.one();
+        return r.getLong("count");
+    }
+
+    public String toString(){
+        String result = "Available Keyspaces and Tables:";
```
mbudiu-vmw, line 253: stringbuilder

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -0,0 +1,264 @@
...
+            s.close();
+            if (sstablePath.isEmpty())
+                throw new RuntimeException("Failed to get SSTablePath");
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        return sstablePath;
+    }
+
+    public Long getRowCount(String keyspace, String table) {
```
mbudiu-vmw, line 246: Where is this used?

platform/src/test/java/org/hillview/test/storage/SSTableTest.java
```diff
@@ -31,7 +33,23 @@
...
-        
+    private final CassandraConnectionInfo info;
+
+    /** This test depends on a local data, I will update this in the future. */
+    public SSTableTest() {
+        String host = "localhost";
+        int jmxPort = 7199;
+        int nativePort = 9042;
+        // The directory where cassandra is installed, check bin/install-cassandra.sh
+        String cassandraRootDir = "/tmp/cassandra";
```
mbudiu-vmw, line 44: make this a final as well in the class.

bin/install-cassandra.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+# This script will install Cassandra at $CASSANDRA_INSTALLATION_DIR/cassandra/ and start it. The process is as follows:
+# 1. Download apache-cassandra-$CASSANDRA_VERSION-bin.tar.gz (if not exist) and then extract it
+# 2. Start Cassandra instance locally
+# 3. Waiting the initialization to complete for 15 seconds
+
+SAVEDIR=$PWD
+# Change this to your preferred installation directory, and make sure to update cassandraRootDir at SSTableTest.java
+CASSANDRA_INSTALLATION_DIR="/tmp"
```
mbudiu-vmw, line 9: Let's move these two variables in bin/lib.sh and make sure you are actually loading lib.sh; look at the other scripts to see how this is done. In this way it's only defined in one place.

data/sstable/loadCassandraDB.sh
```diff
@@ -0,0 +1,26 @@
...
+set -e
+
+CASSANDRA_INSTALLATION_DIR="/tmp"
+CURR_DIR=$PWD
+
+# Prepare the data
+./generate-data.py
+
+# this will append a relative path to load the csv data
+if [[ $(wc -l < cassdb.cql) -le 93 ]]; then
```
mbudiu-vmw, line 14: Why is this necessary?

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -0,0 +1,251 @@
...
+                info.user, info.password);
+    }
+
+    public static class TokenRange {
+        public String start_token;
+        public String end_token;
+        public List<String> endpoints = new ArrayList<>();
+
+        public TokenRange (String rawEntry) {
+            try{
```
mbudiu-vmw, line 130: space is still missing after try; why did you mark the conversation resolved?

data/sstable/loadCassandraDB.sh
daniarherikurniawan: This will append the address of flights_data.csv at the end of cassdb.cql.
We can't put a static/predefined address in the cassdb.cql  because Hillview's root and Cassandra's root are arbitrary.

mbudiu-vmw: Since cassdb.cql is in the same directory can't use just use "."?

mbudiu-vmw: In this case, you don't need to modify the script at all

daniarherikurniawan: Yes, the cassdb.cql is in the same dir, but the flights_data.csv is in Hillview's dir. In the last line of cassdb.cql, the Copy script will look similar to this:

COPY CassDB.flights (ID,DayOfWeek,FlightDate,UniqueCarrier,Origin,OriginCityName,OriginState,Dest,DestState,DepTime,DepDelay,ArrTime,ArrDelay,Cancelled,ActualElapsedTime,Distance)  FROM '/Users/daniar/Documents/GitPR/hillview/data/sstable/flights_data.csv' WITH DELIMITER=',' AND HEADER=TRUE; exit;

PR #638, opened 2020-07-10
More robust cassandra basic tests

mbudiu-vmw:


mbudiu-vmw:
The tests were failing if Cassandra wasn't installed.
We also don't want them to succeed if cassandra is installed but the actual data the tests operate on is wrong.

mbudiu-vmw:
I will actually merge this, because I have broken the build without it, but you should review it anyway.

PR #643, opened 2020-07-13
Add dialog UI to get Cassandra's connection info 

daniarherikurniawan:


web/src/main/webapp/initialObject.ts
```diff
@@ -182,6 +183,24 @@ export class InitialObject extends RemoteObject {
         rr.invoke(observer);
     }
 
+    public loadCassandraFiles(conn: CassandraConnectionInfo, loadMenuPage: FullPage): void {
+        const rr = this.createStreamingRpcRequest<RemoteObjectId>("findCassandraFiles", conn);
+        const files: FileSetDescription = {
+            fileNamePattern: "*-big-Data.db",
+            schemaFile: "schema",
+            headerRow: true,
```
mbudiu-vmw, line 191: You can skip this one.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -205,19 +249,142 @@ public Schema getSchema(CFMetaData metadata) {
...
+                    this.arrConverters.add(CassandraSSTableLoader::convertTimestampToDate);
+                    this.arrConvertedTypes.add(ConvertedType.Date);
+                    break;
+                case "date":
+                case "time":
+                case "blob":
+                    this.arrConverters.add(CassandraSSTableLoader::converterStub);
+                    this.arrConvertedTypes.add(ConvertedType.CQLString);
+                    break;
+                case "counter":
```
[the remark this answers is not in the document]
daniarherikurniawan: Counter is one of Cassandra's type. More about this: https://docs.datastax.com/en/cql-oss/3.x/cql/cql_reference/counter_type.html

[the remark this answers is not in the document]
mbudiu-vmw: seems to be just an integer, treat it like an integer.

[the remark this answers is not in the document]
daniarherikurniawan: Actually it is a Double, the conversion is unusual. Here is the converter :
ByteBuffer counterValue = LongType.instance.decompose(CounterContext.instance().total(((Cell) cd).value()));
Double value = Double.parseDouble(cellType.getSerializer().toCQLLiteral(counterValue));

It must be decomposed first then can be deserialized.
If we convert the byte to double directly, the value will not make sense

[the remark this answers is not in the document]
mbudiu-vmw: How about just this: CounterContext.instance().total(((Cell) cd).value());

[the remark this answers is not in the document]
daniarherikurniawan: It works. Thanks!

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -257,51 +367,89 @@ public ITable load() {
...
+            IAppendableColumn col;
+            // Extracting the value for each column from a single row
+            for (ColumnData cd : row) {
+                // Filter the column to only load the one marked true by columnToLoad[]
+                if (columnToLoad[i]) {
+                    if (cd.column().isSimple()) {
+                        col = columns.get(currentColumn);
+                        AbstractType<?> cellType = cd.column().cellValueType();
+                        value = cellType.getSerializer().deserialize(((Cell) cd).value());
+                        if(value == null) {
```
mbudiu-vmw: missing space

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -257,51 +367,89 @@ public ITable load() {
...
+                            col.appendMissing();
+                        } else {
+                            switch (columTypes.get(i)) {
+                                case STRING:
+                                    col.append(value.toString());
+                                    break;
+                                case CQLSTRING:
+                                    col.append(cellType.getSerializer()
```
mbudiu-vmw: how is this different from just value.toString()?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -257,51 +367,89 @@ public ITable load() {
...
+                                    col.append(((Byte) value).intValue());
+                                    break;
+                                case VARINT:
+                                    col.append(((BigInteger) value).doubleValue());
+                                    break;
+                                case BIGINT:
+                                    col.append(((Long) value).doubleValue());
+                                    break;
+                                case DECIMAL:
+                                    col.append(((BigDecimal) value).doubleValue());
```
mbudiu-vmw: add a helper function which does bounds checking and throws if the value is too large

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: When it comes to converting time and date, the result is not as expected. For example:
- Original Cassandra format
        date : 2020-07-14
        time : 13:30:23.123000000
- Use normal serializer 
        date : -2147465191    
        time : 48623123000000
- Use serializer.toCQLLiteral
        date : 2020-07-14
        time : 13:30:23.123000000

That's why I create a new type as CQLSTRING

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -58,13 +65,31 @@
...
     private final Schema actualSchema;
     private final String ssTablePath;
+    private long rowCount;
 
     public boolean lazyLoading;
     public final CFMetaData metadata;
     private final SSTableReader ssTableReader;
+    // private List<TypeSerializer<Object>> arrTypeSerializers = new ArrayList<>();
+
+    enum CassandraType {
```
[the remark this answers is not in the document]
daniarherikurniawan: I remove the enum, and now the latency is 10% higher. I will commit the update soon.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -45,26 +47,33 @@
...
+import org.apache.commons.lang3.Range;
 
 import javax.annotation.Nullable;
 import java.util.stream.Stream;
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.util.*;
 import java.util.stream.StreamSupport;
 
+import static org.junit.Assert.assertTrue;
```
mbudiu-vmw, line 59: do not use junit in your code, only in tests.

platform/src/main/java/org/hillview/maps/FindCassandraFilesMap.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    public List<IFileReference> apply(@Nullable Empty empty) {
+        CassandraDatabase db = new CassandraDatabase(this.conn);
+        List<String> ssTables = db.getSSTablePath();
+        db.closeClusterConnection();
+        return Linq.map(ssTables, CassandraFileReference::new);
+    }
+
```
mbudiu-vmw, line 47: still an extra empty line

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -115,52 +124,61 @@ public CFMetaData getSSTableMetadata(Descriptor desc) throws Exception{
...
         ContentsKind kind;
-        switch (colType) {
-            case "boolean":
-            case "ascii":
-            case "inet":
-            case "text":
-            case "timeuuid":
-            case "uuid":
+        CQL3Type type = colDef.type.asCQL3Type();
+        assertTrue(type instanceof CQL3Type.Native);
```
mbudiu-vmw, line 129: instanceof may be expensive, perhaps it's better not to test this, the next cast will throw if that does not work.

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -136,24 +165,64 @@ public TokenRange (String rawEntry) {
...
+        return IPAddress;
+    }
+
+    private String discoverLocalEndpoint() throws Exception {
+        Path cassandraPath = Paths.get(this.info.cassandraRootDir);
+        Path cassandraYaml = Paths.get(cassandraPath.toString(), CassandraDatabase.cassandraYamlPath);
+        String nodeName = null;
+        File cassandraConfig = cassandraYaml.toFile();
+        Scanner myReader = new Scanner(cassandraConfig);
+        while (myReader.hasNextLine()) {
```
mbudiu-vmw, line 192: isn't there a library to read yaml so you don't need to do this in this way?

daniarherikurniawan: There is, but it will load all config parameters. We just need one parameter in the beginning, so I was thinking to get it manually to save space and time. For now, I will update this to use yaml-reader library

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,17 +235,78 @@ public Schema getSchema(CFMetaData metadata) {
         }
     }
 
+    public Schema getSchema() {
+        return this.actualSchema;
+    }
+
+    private static SSTableReadsListener newReadCountUpdater() {
```
mbudiu-vmw, line 242: why is this needed?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,17 +235,78 @@ public Schema getSchema(CFMetaData metadata) {
...
+                // No need to do sstable.incrementReadCount();
+            }
+        };
+    }
+
+    private void computeRowCount() {
+        this.rowCount = 0;
+        PartitionColumns newColDef = this.columnDefinitions;
+        for (ColumnDefinition columnDefinition : this.columnDefinitions)
+            newColDef = newColDef.without(columnDefinition);
```
mbudiu-vmw, line 255: There must be a way to create empty column definitions instead of adding all and removing them.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -58,44 +68,90 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * This is to support loading local sstable test without a token-range. The
+     * generated token range will cover all partition in a single range.
+     */
+    @VisibleForTesting
+    public CassandraSSTableLoader(String ssTablePath, boolean lazyLoading) {
```
mbudiu-vmw, line 116: Do not duplicate code, instead call `this(...)`.

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -136,24 +165,64 @@ public TokenRange (String rawEntry) {
...
-            return "start: " + this.start_token + ", end: " + this.end_token +
-                ", endpoints: " + this.endpoints.toString();
+            StringBuilder sb = new StringBuilder();
+            sb.append("tokenRange: " + this.tokenRange.toString());
+            sb.append(", endpoints: " + this.endpoints.toString());
+            return sb.toString();
+        }
+    }
+
+    private String convertEndpointToIP(String endpoint) {
```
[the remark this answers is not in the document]
daniarherikurniawan: If I do not convert that to an IP, when I input `cass-1.cassandra-mitmem.cs331-uc.emulab.net`  I got this exception:
```
java.lang.Exception: java.lang.RuntimeException: java.lang.RuntimeException: java.net.UnknownHostException: http://cass-1.cassandra-mitmem.cs331-uc.emulab.net/: Name or service not known
	at org.hillview.dataset.LocalDataSet.lambda$flatMap$1(LocalDataSet.java:128)
Caused by: java.lang.RuntimeException: java.lang.RuntimeException: java.net.UnknownHostException: http://cass-1.cassandra-mitmem.cs331-uc.emulab.net/: Name or service not known
	at org.hillview.storage.CassandraDatabase.<init>(CassandraDatabase.java:100)
	
```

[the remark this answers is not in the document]
mbudiu-vmw: why is the https there?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -58,44 +68,90 @@
...
     public boolean lazyLoading;
     public final CFMetaData metadata;
     private final SSTableReader ssTableReader;
 
-    static{
-        // Initializing DatabaseDescriptor that contains Cassandra's general configuration
+    private final PartitionColumns columnDefinitions;
+    private final IPartitioner partitioner;
+    private final List<CassandraTokenRange> tokenRanges;
+    private static String localEndpoint;
```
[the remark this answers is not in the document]
daniarherikurniawan: Because it will be used by `private static void loadColumns()`

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -203,105 +327,154 @@ public Schema getSchema(CFMetaData metadata) {
         @Override
         public List<? extends IColumn> loadColumns(List<String> names) {
             boolean[] columnToLoad = this.getColumnMarker(names);
-            ISSTableScanner currentScanner = this.ssTableReader.getScanner();
-            Spliterator<UnfilteredRowIterator> splititer = Spliterators.spliteratorUnknownSize(currentScanner,
-                Spliterator.IMMUTABLE);
-            Stream<UnfilteredRowIterator> partitions = StreamSupport.stream(splititer, false);
+            List<IAppendableColumn> columns = createColumns(columnToLoad, this.columnCountToLoad);
```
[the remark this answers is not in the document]
daniarherikurniawan: Java doesn't allow me to do that, because this call is from inside an inner class ` SSTableColumnLoader`

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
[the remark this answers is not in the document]
daniarherikurniawan: I believe the UI might add `http` somehow

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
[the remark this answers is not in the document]
mbudiu-vmw: You can with the syntax `SSTableColumnLoader.this.createColumn`.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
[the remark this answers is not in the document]
mbudiu-vmw: Don't make it static, pass it as an argument to the function.
As I said, statics are really bad for multithreaded execution.
Many of these functions may run simultaneously on multiple cores.

web/src/main/webapp/initialObject.ts
```diff
@@ -202,4 +217,12 @@ export class InitialObject extends RemoteObject {
             { kind: "DB", description: conn }, "loading database table");
         rr.invoke(observer);
     }
+
+    public loadLocalDBTable(jdbcConn: JdbcConnectionInformation, cassConn: CassandraConnectionInfo, loadMenuPage: FullPage): void {
+        if (jdbcConn.databaseKind == "cassandra") {
```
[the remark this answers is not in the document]
daniarherikurniawan: I believe I shouldn't remove this function.
The dialog for "Local DB table..." is the same as "Federated DB tables...", both are displaying `DBDialog()`. So, there is always an option to load Cassandra from any of both menus. Should I create a similar method of `loadLocalDBTable` to handle the federated databases? I will call it `loadFederatedDBTable`
Or, should I differentiate the menu for loading local table vs federated table?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: To satisfy the input arguments of `ssTableReader.getScanner`. We need to filter the data using  `DataRange` and `ColumnFilter` when getting the scanner. Although there are several types of `getScanner`, but the one with both filters requires `SSTableReadsListener` as the last input argument.

web/src/main/webapp/initialObject.ts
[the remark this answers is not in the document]
mbudiu-vmw: "Local" usually means that you connect using jdbc and read from a single node. "Federated" means that you read in parallel from multiple partitions. For Cassandra we want the latter. The "local" option is mostly useful for various experiments.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: I can't use `this`, because there is `super` initialization. I ended up creating a static method to return CassandraSSTableLoader object

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -115,53 +170,62 @@ public CFMetaData getSSTableMetadata(Descriptor desc) throws Exception{
...
+        switch ((CQL3Type.Native) colDef.type.asCQL3Type()) {
+            case BOOLEAN:
+            case ASCII:
+            case INET:
+            case TEXT:
+            case TIMEUUID:
+            case UUID:
+            case DURATION:
+            case TIME:
+            case DATE:
```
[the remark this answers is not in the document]
daniarherikurniawan: I've tried that, but the format of `Date, Time, and Duration` at Cassandra can't be converted directly to Java style. We must get the formating style before converting it. Especially, the formatting of Cassandra's `Duration` is unlike Java's duration, I have to convert, split, and parse the data to match Java's Duration.  It will be faster if client applies `filter` to these fields as needed. What do you think?

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -105,29 +125,38 @@ public void closeClusterConnection() {
 
     /**
      * Connection for running server-side query.
-    */
+     */
     private void connectLocalProbe() throws IOException {
         Converters.checkNull(this.nodeProbeFactory);
         if (Utilities.isNullOrEmpty(info.user))
-            this.probe = this.nodeProbeFactory.create(info.host, info.jmxPort);
+            this.probe = this.nodeProbeFactory.create("127.0.0.1", info.jmxPort);
```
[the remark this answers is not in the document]
daniarherikurniawan: fixed!

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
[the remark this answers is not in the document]
daniarherikurniawan: Turns out, there was an auto-complete at my browser to add HTTP when copy-pasting a URL. I removed this `convertEndpointToIP`

web/src/main/webapp/initialObject.ts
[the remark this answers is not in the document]
daniarherikurniawan: I changed loadLocalDBTable to `loadFederatedDBTable`, and Cassandra will only appear as an option if you click `Federated DB` on the UI

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -181,67 +236,60 @@ private void setStoredTableInfo() {
         for (KeyspaceMetadata kMetadata : keyspaces) {
             keyspace = kMetadata.getName();
             tables = this.metadata.getKeyspace(keyspace).getTables();
-            tableNames = tables.stream().map(AbstractTableMetadata::getName)
-                .collect(Collectors.toList());
+            tableNames = tables.stream().map(AbstractTableMetadata::getName).collect(Collectors.toList());
```
mbudiu-vmw, line 239: Move the declaration of the variables here; they do not need to exist outside the loop.

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -181,67 +236,60 @@ private void setStoredTableInfo() {
...
-        Path cassandraPath = Paths.get(info.cassandraRootDir);
+     * Given the keyspace and table name, sstableutil will return the
+     * sstable-related files. Using sstableutil command is better than doing
+     * manually sstable search. Although we know the sstable parent dir from
+     * cassandra.yaml, we need to rule out the outdated sstables. And sstableutil is
+     * built specifically to identify which version are the most updated.
+     */
+    public List<String> getSSTablePath() {
+        List<String> result = new ArrayList<>();
+        boolean isWindows = System.getProperty("os.name").toLowerCase().startsWith("windows");
```
mbudiu-vmw, line 258: Please move this line into a separate static function in the class Utilities. Call it "runningOnWindows()".

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -58,44 +76,75 @@
...
+            this.actualSchema = this.convertSchema();
+            this.computeRowCount();
+        } catch (Exception e) {
+            HillviewLogger.instance.error("Failed initializing Metadata and SStable partitions", "{0}",
+                    this.ssTablePath);
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
```
mbudiu-vmw, line 119: Please add a comment that this function is only used for testing.
The function name should also reflect that: getTestTableLoader?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
     }
 
-    public long getNumRows() {
-        ISSTableScanner currentScanner = this.ssTableReader.getScanner();
-        Spliterator<UnfilteredRowIterator> splititer = Spliterators.spliteratorUnknownSize(currentScanner,
-            Spliterator.IMMUTABLE);
-        Stream<UnfilteredRowIterator> partitions = StreamSupport.stream(splititer, false);
-        return partitions.count();
+    public int getRowCount() {
+        if (this.rowCount > Integer.MAX_VALUE || this.rowCount < Integer.MIN_VALUE)
```
mbudiu-vmw, line 267: Use Converters.toInt, which already makes this check.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+     * uses boolean marker stored at columnToLoad to recognize the needed columns
+     */
+    private List<IAppendableColumn> createColumns(List<String> names) {
         List<ColumnDescription> cols = Converters.checkNull(this.actualSchema).getColumnDescriptions();
-        List<IAppendableColumn> result = new ArrayList<IAppendableColumn>(columnCountToLoad);
-        int i = 0;
+        List<IAppendableColumn> result = new ArrayList<IAppendableColumn>(names.size());
         for (ColumnDescription cd : cols) {
-            if (columnToLoad[i])
+            if (names.contains(cd.name))
```
mbudiu-vmw, line 280: This could a quadratic cost, which could be large for more than 1000 columns.
Create a set from the List of names.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
-                        if (columnToLoad[i]) {
-                            if (cd.column().isSimple()) {
-                                Cell cell = (Cell) cd;
-                                AbstractType<?> cellType = cell.column().cellValueType();
-                                value = cellType.getSerializer().deserialize(cell.value());
-                            } else {
-                                // Hillview won't process the complex data
-                                throw new RuntimeException("Hillview can't convert complex data");
+    /**
+     * Lazy and non-lazy loader will call this function to load specific column
```
mbudiu-vmw, line 344: In general you don't want to document here who calls a specific function. This kind of comment is hard to maintain.
But you want to describe what the function does.

mbudiu-vmw: Note that columnToLoad no longer exists, so this comment is already inaccurate.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+     */
+    private static void loadColumns(SSTableReader ssTableReader, List<IAppendableColumn> columns,
+            PartitionColumns columnDefinitions, List<CassandraTokenRange> tokenRanges, String localEndpoint) {
+        int colCount = columnDefinitions.size();
+        ColumnFilter cf = ColumnFilter.selection(columnDefinitions);
+        List<Long> arrPrefixComparison = new ArrayList<>();
+        HashMap<CQL3Type, TypeSerializer<?>> serializers = new HashMap<CQL3Type, TypeSerializer<?>>();
+        Iterator<ColumnDefinition> colIter = columnDefinitions.regulars.simpleColumns();
+        // init reusable serializer and prefixComparison
+        while(colIter.hasNext()) {
```
mbudiu-vmw, line 355: no space.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+                            Row row = (Row) unfiltered;
+                            int currentColumn = 0;
+                            Object value;
+                            IAppendableColumn col;
+                            // Extracting the value of each column from a single row
+                            for (ColumnData cd : row) {
+                                col = columns.get(currentColumn);
+                                // Missing column case #1 
+                                // Where unloaded missing column(s) is between idx 0 until last non missing column
+                                while (cd.column().name.prefixComparison != arrPrefixComparison
```
mbudiu-vmw, line 382: Do not put a loop in the innermost loop, this will make your code many times slower.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+    /**
+     * Lazy and non-lazy loader will call this function to load specific column
+     * marked by columnToLoad
+     */
+    private static void loadColumns(SSTableReader ssTableReader, List<IAppendableColumn> columns,
+            PartitionColumns columnDefinitions, List<CassandraTokenRange> tokenRanges, String localEndpoint) {
+        int colCount = columnDefinitions.size();
+        ColumnFilter cf = ColumnFilter.selection(columnDefinitions);
+        List<Long> arrPrefixComparison = new ArrayList<>();
+        HashMap<CQL3Type, TypeSerializer<?>> serializers = new HashMap<CQL3Type, TypeSerializer<?>>();
```
mbudiu-vmw, line 352: use an array, not a hashmap, it's much faster.
A raw array, not an ArrayList.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+        while(colIter.hasNext()) {
+            ColumnDefinition cd = colIter.next();
+            CQL3Type type = cd.type.asCQL3Type();
+            serializers.put(type, type.getType().getSerializer());
+            arrPrefixComparison.add(cd.name.prefixComparison);
+        }
+
+        for (CassandraTokenRange tr : tokenRanges) {
+            // Load the partition if the local endpoint is listed as the first endpoint
+            if (tr.endpoints.get(0).equals(localEndpoint)) {
```
mbudiu-vmw, line 364: to make the code clearer for such long bodies it's simpler to write the converse:
`if (!tr.endpoinds...) continue;`
This will have one fewer levels of indentation.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
-                                AbstractType<?> cellType = cell.column().cellValueType();
-                                value = cellType.getSerializer().deserialize(cell.value());
-                            } else {
-                                // Hillview won't process the complex data
-                                throw new RuntimeException("Hillview can't convert complex data");
+    /**
+     * Lazy and non-lazy loader will call this function to load specific column
+     * marked by columnToLoad
+     */
+    private static void loadColumns(SSTableReader ssTableReader, List<IAppendableColumn> columns,
```
mbudiu-vmw, line 347: Since you access columns in the innermost look it may be faster to make it a raw array.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+                        if (unfiltered instanceof Row) {
+                            Row row = (Row) unfiltered;
+                            int currentColumn = 0;
+                            Object value;
+                            IAppendableColumn col;
+                            // Extracting the value of each column from a single row
+                            for (ColumnData cd : row) {
+                                col = columns.get(currentColumn);
+                                // Missing column case #1 
+                                // Where unloaded missing column(s) is between idx 0 until last non missing column
```
mbudiu-vmw, line 381: What is an unloaded missing column?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+                                while (cd.column().name.prefixComparison != arrPrefixComparison
+                                        .get(currentColumn)) {
+                                    // The column with null value will not be included, so here 
+                                    // we add the null identifier for the skipped column(s)
+                                    col.appendMissing();
+                                    currentColumn++;
+                                    col = columns.get(currentColumn);
+                                }
+                                CQL3Type colType = cd.column().type.asCQL3Type();
```
mbudiu-vmw, line 390: you don't need this, you already have them in the array you already built before.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+                SSTableReadsListener listener = CassandraSSTableLoader.newReadCountUpdater();
+                ISSTableScanner currentScanner = ssTableReader.getScanner(cf, range, false, listener);
+
+                Spliterators.spliteratorUnknownSize(currentScanner, Spliterator.CONCURRENT)
+                    .forEachRemaining((partition) -> {
+                        Unfiltered unfiltered = partition.next();
+                        if (unfiltered instanceof Row) {
+                            Row row = (Row) unfiltered;
+                            int currentColumn = 0;
+                            Object value;
```
mbudiu-vmw, line 375: you don't need this declaration here.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+                                    currentColumn++;
+                                    col = columns.get(currentColumn);
+                                }
+                                CQL3Type colType = cd.column().type.asCQL3Type();
+                                ByteBuffer byteBuff = ((Cell) cd).value();
+                                value = serializers.get(colType).deserialize(byteBuff);
+                                // Missing column case #2
+                                // Where loaded missing column(s) is still loaded and the capacity is 0
+                                if (byteBuff.capacity() == 0) {
+                                    col.appendMissing();
```
mbudiu-vmw, line 396: isn't value null in this case?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+                    });
+                ;
             }
-        });
-        return columns;
+        }
+    }
+
+    private static boolean bigDecimalInRange(BigDecimal bigDecimal) {
+        org.apache.commons.lang3.Range<BigDecimal> doubleRange = org.apache.commons.lang3.Range
```
mbudiu-vmw, line 491: This is extremely expensive: you are allocating an object for every row.
Also, make this function return the double rather than a boolean and throw otherwise.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+                                            Duration duration = DurationSerializer.instance.deserialize(byteBuff);
+                                            if (duration.getMonths() == 0) {
+                                                int days = duration.getDays();
+                                                long nanos = duration.getNanoseconds();
+                                                long millis = (nanos + Duration.NANOS_PER_HOUR * 24 * days) / 1000000;
+                                                col.append(Converters.toDuration(millis));
+                                            } else {
+                                                // java.time.Duration support day and time (but not month and year)
+                                                throw new RuntimeException(
+                                                        "Failed to convert Duration with months != 0");
```
mbudiu-vmw, line 465: The message should rather say that Cassandra Durations with months are not supported.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+                Spliterators.spliteratorUnknownSize(currentScanner, Spliterator.CONCURRENT)
+                    .forEachRemaining((partition) -> {
+                        Unfiltered unfiltered = partition.next();
+                        if (unfiltered instanceof Row) {
+                            Row row = (Row) unfiltered;
+                            int currentColumn = 0;
+                            Object value;
+                            IAppendableColumn col;
+                            // Extracting the value of each column from a single row
+                            for (ColumnData cd : row) {
```
mbudiu-vmw, line 378: instead of iterating over the row, iterate over ColumnDefinition and call row.getCell(col)

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+                                            break;
+                                        case FLOAT:
+                                            col.append(((Float) value).doubleValue());
+                                            break;
+                                        case COUNTER:
+                                            col.append(CounterContext.instance().total(byteBuff));
+                                            break;
+                                        case TIME:
+                                            // TODO: Convert Time type to Java Time instead of Java Instant
+                                            long myTime = TimeSerializer.instance.deserialize(byteBuff);
```
mbudiu-vmw, line 441: You already have the cassandra Time in the value. Why do you need to do this?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+                                                col.append(((BigDecimal) value).doubleValue());
+                                            break;
+                                        case DOUBLE:
+                                            col.append((Double) value);
+                                            break;
+                                        case FLOAT:
+                                            col.append(((Float) value).doubleValue());
+                                            break;
+                                        case COUNTER:
+                                            col.append(CounterContext.instance().total(byteBuff));
```
mbudiu-vmw, line 437: Isn't value already the long you need?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+                                            // Downgrade to millis and append 2000-1-1 as the date
+                                            myTime = myTime / 1000000 + 946706400000L;
+                                            col.append(Instant.ofEpochMilli(myTime));
+                                            break;
+                                        case TIMESTAMP:
+                                            col.append(
+                                                    TimestampSerializer.instance.deserialize(byteBuff).toInstant());
+                                            break;
+                                        case DATE:
+                                            long msTime = SimpleDateSerializer
```
mbudiu-vmw, line 451: again, you already have value which has type Date.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +230,269 @@ public Schema getSchema(CFMetaData metadata) {
...
+                                                long nanos = duration.getNanoseconds();
+                                                long millis = (nanos + Duration.NANOS_PER_HOUR * 24 * days) / 1000000;
+                                                col.append(Converters.toDuration(millis));
+                                            } else {
+                                                // java.time.Duration support day and time (but not month and year)
+                                                throw new RuntimeException(
+                                                        "Failed to convert Duration with months != 0");
+                                            }
+                                            break;
+                                        case EMPTY:
```
mbudiu-vmw, line 468: For EMPTY can't you just use a None column type?
In that case you don't have to do anything here.

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -27,57 +30,74 @@
...
-
-    /** This class enables Hillview to execute server-side and client-side queries, such as
-     * getting sstable path, list of stored tables, list of token-range, and also the row count
-     * of a specific table.
-     * The flow of execution to locate a set of table partitions is as follows:
-     * - CassandraDatabase establishes NodeProbe connection to get full token-range and its endpoints
-     * - CassandraDatabase find the sstable path by running sstableutil
-     * - Finally, CassandraSSTableReader uses both output to read the local token-range
-    */
+    private final List<CassandraTokenRange> tokenRanges = new ArrayList<CassandraTokenRange>();
```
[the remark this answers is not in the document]
daniarherikurniawan: I move this initialization to the constructor

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: But, I need to convert it back to List at the end because of this override:
```
public interface IColumnLoader {
List<? extends IColumn> loadColumns(List<String> names);
}
```
Converting it back will not add significant overhead, right?

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: To get the Long representation, because you were asking me to add 00-01-01 as the date. There is no method at Cassandra TimeType to get the Long representation directly, thus I deserialize the byteBuff. Please let me know if I misunderstood anything.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: Although the type is also Long, but Value isn't the right value. The (Long) Value gives me 422215120878763L, but the current value is 1L.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: Not always null, for example when the type is BigInt.

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -27,57 +30,75 @@
...
 import org.apache.cassandra.tools.INodeProbeFactory;
 import org.apache.cassandra.tools.NodeProbe;
 import org.apache.cassandra.tools.NodeProbeFactory;
+import org.apache.cassandra.utils.FBUtilities;
 import org.hillview.utils.Converters;
 import org.hillview.utils.HillviewLogger;
 import org.hillview.utils.Utilities;
-
-import static org.apache.commons.lang3.StringUtils.EMPTY;
+import org.yaml.snakeyaml.Yaml;
```
mbudiu-vmw, line 46: doesn't this require a new dependence in pom.xml?

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -136,22 +168,45 @@ public TokenRange (String rawEntry) {
...
     }
 
+    private String discoverLocalEndpoint() throws Exception {
+        Path cassandraPath = Paths.get(this.info.cassandraRootDir);
+        Path cassandraYaml = Paths.get(cassandraPath.toString(), CassandraDatabase.cassandraYamlPath);
+        Yaml yamlConfig = new Yaml();
+        InputStream is = new FileInputStream(cassandraYaml.toFile());
+        Config config = yamlConfig.loadAs(is, Config.class);
+        InetAddress inet = InetAddress.getByName(config.listen_address);
+        return inet.getHostAddress().toString();
```
mbudiu-vmw, line 185: I wonder whether this is ever something different that localhost.

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
```diff
@@ -163,85 +218,79 @@ public CassTable(String keyspace, List<String> tables) {
...
-     * Using sstableutil command is better than doing manually sstable search. Although we know
-     * the sstable parent dir from cassandra.yaml, we need to rule out the outdated sstables.
-     * And sstableutil is built specifically to identify which version are the most updated.
-      */
-    public String getSSTablePath() {
-        boolean isWindows = System.getProperty("os.name")
-            .toLowerCase().startsWith("windows");
-        Path ssTableUtilPath = Paths.get(ssTableUtilDir);
-        Path cassandraPath = Paths.get(info.cassandraRootDir);
+     * Given the keyspace and table name, sstableutil will return the
```
mbudiu-vmw, line 251: in general the JavaDoc documentation of a method should just say *what* the method does, and not *how* it does it.
This kind of of comment is more suitable for the body of the method.

platform/src/main/java/org/hillview/utils/Utilities.java
```diff
@@ -176,6 +176,14 @@ public static String getHostName() {
         }
     }
 
+    /**
+     * Checks whether the local OS is Windows
+     *
```
mbudiu-vmw, line 181: no need for this extra empty line. The comment should say: return true if the operating system of the executing program is Windows.

platform/src/test/java/org/hillview/test/storage/SSTableTest.java
```diff
@@ -136,6 +142,24 @@ public void testCassandraDirectory() {
...
+            // Connecting to Cassandra node and get some data
+            conn = this.getConnectionInfo();
+            db = new CassandraDatabase(conn);
+        } catch (Exception e) {
+            // this will fail if local Cassandra is not running, but we don't want to fail the test.
+            this.ignoringException("Failed connecting to local cassandra", e);
+            return;
+        }
+        List<CassTable> storedTable = db.getStoredTableInfo();
+        assert(storedTable.toString().contains("cassdb: [ test counter flights users]"));
```
mbudiu-vmw, line 160: Do not use assert in tests, use the junit Assert.assertTrue.

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -171,137 +232,236 @@ public Schema getSchema(CFMetaData metadata) {
...
+        for (CassandraTokenRange tr : tokenRanges) {
+            // Load the partition if the local endpoint is listed as the first endpoint
+            if (!tr.endpoints.get(0).equals(localEndpoint)) continue;
+            DataRange range = DataRange.forTokenRange(tr.tokenRange);
+            SSTableReadsListener listener = CassandraSSTableLoader.newReadCountUpdater();
+            ISSTableScanner currentScanner = ssTableReader.getScanner(cf, range, false, listener);
+            Spliterators.spliteratorUnknownSize(currentScanner, Spliterator.CONCURRENT)
+                .forEachRemaining((partition) -> {
+                    Unfiltered unfiltered = partition.next();
+                    if (unfiltered instanceof Row) {
```
mbudiu-vmw, line 368: Could you assert here that unfiltered is an instance of Row?
What else could it be?

platform/src/main/java/org/hillview/storage/CassandraDatabase.java
daniarherikurniawan: Yes, when I deploy Cassandra to the cluster, I need to change the `listen_address` to the node IP

platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
daniarherikurniawan: The other possibility is to have `RangeTombstoneMarker`. This is the original Cassandra's code to read the sstable:
```
Unfiltered unfiltered;
while (partition.hasNext())
{
    unfiltered = partition.next();
    if (unfiltered instanceof Row)
    {
        serializeRow((Row) unfiltered);
    }
    else if (unfiltered instanceof RangeTombstoneMarker)
    {
        serializeTombstone((RangeTombstoneMarker) unfiltered);
    }
    updatePosition();
}
```

PR #650, opened 2020-07-23
Expandable chart menu

daniarherikurniawan:
This PR adds two interfaces to create a submenu:
1. this.contextMenu.addExpandableItem
This is similar to addItem, the code will only add `expandableMenu` as an additional class to that item.
2. this.contextMenu.insertSubMenu(parentMenu, mi, enabled)
This will create a submenu item of the parentMenu.

### DIFF
--- a/web/src/main/webapp/dataViews/tableView.ts
+++ b/web/src/main/webapp/dataViews/tableView.ts
@@ -640,46 +640,6 @@ export class TableView extends TSViewBase implements IScrollTarget, OnNextK {
               },
               true
             );
-            this.contextMenu.addExpandableItem(
-              {
-                text: "More...",
-                action: () => this.contextMenu.expandMenu("More..."),
-                help: "List of available charts to draw. " + "",
-              },
-              true
-            );
-            this.contextMenu.addItem({
-                text: "Histogram",
-                action: () => this.chart(this.schema.getDescriptions(this.getSelectedColNames()),
-                    this.getSelectedColCount() === 1 ? "Histogram" : "2DHistogram"),
-                help: "Plot the data in the selected columns as a histogram. " +
-                    "Applies to one or two columns only.",
-            }, selectedCount >= 1 && selectedCount <= 2);
-            this.contextMenu.addItem({
-                text: "Quartile vector",
-                action: () => this.chart(this.schema.getDescriptions(this.getSelectedColNames()), "QuartileVector"),
-                help: "Plot the data in the selected columns as a vector of quartiles. " +
-                    "Applies to one or two columns only.",
-            }, selectedCount == 2);
-            this.contextMenu.addItem({
-                text: "Heatmap",
-                action: () => this.chart(this.schema.getDescriptions(this.getSelectedColNames()), "Heatmap"),
-                help: "Plot the data in the selected columns as a heatmap. " +
-                    "Applies to two columns only.",
-            }, selectedCount === 2);
-            this.contextMenu.addItem({
-                text: "Trellis histograms",
-                action: () => this.chart(this.schema.getDescriptions(this.getSelectedColNames()),
-                    selectedCount > 2 ? "Trellis2DHistogram" : "TrellisHistogram"),
-                help: "Plot the data in the selected columns as a Trellis plot of histograms. " +
-                    "Applies to two or three columns only.",
-            }, selectedCount >= 2 && selectedCount <= 3);
-            this.contextMenu.addItem({
-                text: "Trellis heatmaps",
-                action: () => this.chart(this.schema.getDescriptions(this.getSelectedColNames()), "TrellisHeatmap"),
-                help: "Plot the data in the selected columns as a Trellis plot of heatmaps. " +
-                    "Applies to three columns only.",
-            }, selectedCount === 3 && !this.isPrivate());
             this.contextMenu.addItem({
                 text: "Rename...",
                 action: () => this.renameColumn(),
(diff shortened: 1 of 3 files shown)


web/src/main/webapp/ui/menu.ts
```diff
@@ -1,3 +1,4 @@
+import { tickStep } from "d3-array";
```
mbudiu-vmw, line 1: what is this needed for?

web/src/main/webapp/hillview.css
```diff
@@ -355,8 +356,8 @@ textarea:focus, input:focus {
 }
 
 table.menu {
-    box-shadow: 4px 4px 4px rgba(0, 0, 0, 0.5);
-    border: 1px solid grey;
+    /* box-shadow: 4px 4px 4px rgba(0, 0, 0, 0.5); */
```
mbudiu-vmw, line 359: Why are you changing these?

web/src/main/webapp/ui/menu.ts
```diff
@@ -40,157 +41,279 @@ export interface MenuItem extends BaseMenuItem {
...
-        return -1;
+function emptySubMenu(): subMenuCells {
+    return {
+        cells: []
     }
+}
 
-    public getHTMLRepresentation(): HTMLElement {
-        return this.outer;
+abstract class BaseMenu<MI extends BaseMenuItem> implements IHtmlElement {
```
mbudiu-vmw, line 54: please use the same formatting as everywhere else: four spaces.
Reformatting this makes is very difficult to compare the code.

web/src/main/webapp/ui/menu.ts
```diff
@@ -249,10 +372,34 @@ export class ContextMenu extends BaseMenu<MenuItem> implements IHtmlElement {
     public setAction(mi: MenuItem, enabled: boolean): void {
         const index = this.find(mi.text);
         const cell = this.cells[index];
-        if (mi.action != null && enabled)
-            cell.onclick = () => { this.hide(); mi.action(); };
-        else
+        if (mi.action != null && enabled) {
+            if (cell.classList.contains("expandableMenu")) {
+              cell.onclick = () => {
```
mbudiu-vmw, line 377: I thought you said that this is hover. Where is the hover check?

web/src/main/webapp/dataViews/tableView.ts
```diff
@@ -744,6 +720,80 @@ export class TableView extends TSViewBase implements IScrollTarget, OnNextK {
                         },
                         help: "Extract a value associated with a specific key."
                     }, !this.isPrivate());
+            this.contextMenu.insertSubMenu( "Charts...",
```
mbudiu-vmw, line 723: Instead of specifying here the text to search for, it would be better for the method addExpandableMenuItem to return a more useful object which has methods to add new items.

web/src/main/webapp/hillview.css
```diff
@@ -107,6 +107,7 @@ table {
     background: linear-gradient(#f9f9f9,#e4e4e4);
     border: 1px solid black;
     font-weight: bold;
+    padding: 2px 5px;
```
mbudiu-vmw, line 110: If you want to change the style you will have to justify these changes.

web/src/main/webapp/ui/menu.ts
```diff
@@ -130,6 +160,56 @@ abstract class BaseMenu<MI extends BaseMenuItem> implements IHtmlElement {
         return cell;
     }
 
+    /**
+     * Inserting subMenu must be done in the end of all parent menu insert. If the
+     * submenu exceed the length of all main menu, we need to add dummyMenu
```
mbudiu-vmw, line 165: I don't understand the role of dummyMenu, can you explain it?

web/src/main/webapp/ui/menu.ts
```diff
@@ -130,6 +160,56 @@ abstract class BaseMenu<MI extends BaseMenuItem> implements IHtmlElement {
...
+        else cell.textContent = mi.text;
+
+        var subIndex = this.subMenuCells[parentIndex].cells.length;
+        this.subMenuCells[parentIndex].cells.push(cell);
+        cell.onmouseenter = () => this.selectSubMenu(parentIndex, subIndex);
+        cell.onmouseleave = () => this.selectSubMenu(parentIndex, -1);
+        this.setSubMenuAction(parentIndex, subMenuIndex, mi, enabled);
+    }
+
+    public addExpandableItem(mi: MI, enabled: boolean): HTMLTableDataCellElement {
```
mbudiu-vmw, line 199: I noticed that when the popup menu is displayed the menu items below all shift down. So there is something wrong with the styling.

web/src/main/webapp/ui/menu.ts
mbudiu-vmw: I see the hover handled with onmouseenter. But why is this necessary?

web/src/main/webapp/ui/menu.ts
```diff
@@ -282,6 +431,11 @@ export class SubMenu extends BaseMenu<MenuItem> implements IHtmlElement {
         else
             cell.onclick = (e: MouseEvent) => { e.stopPropagation(); this.hide(); };
     }
+
+    public setSubMenuAction(parentIndex: number, subMenuIndex: number,
```
mbudiu-vmw, line 435: rather than have this method be abstract, it's better to make the base method throw.

web/src/main/webapp/ui/menu.ts
mbudiu-vmw: If I click on the menu item it does not disappear; in fact, the whole menu stays resident after the action, and there seems to be no way to remove the menu.

web/src/main/webapp/hillview.css
daniarherikurniawan: The shadow and border make the menu appears to be shifting down when you expand the menu. 
image

That image should explain how these submenus are displayed. A dummyMenu is needed when the most left column is no longer has a menu. The dummyMenu makes a submenu stay on the right column.  I displayed dummyMenu as a transparent menu. 

When a Submenu is not displayed, the display will be `none`, thus other submenus can take the space and appear to be displayed in the second column as depicted in the next image.
image

web/src/main/webapp/ui/menu.ts
daniarherikurniawan: Ideally, we want the `expandableMenu` to expand the submenus only, without having another onclick() task. The main purpose of this code is to prevent `this.hide();` because the menu should stay still while the submenus are being displayed. Do you agree if I remove the setAction for `expandableMenu`? The onmouseenter will display the submenus

web/src/main/webapp/hillview.css
daniarherikurniawan: The padding is intended to separate the label from the column border. 
image
Should I undo this?

web/src/main/webapp/ui/menu.ts
daniarherikurniawan: I put the explanation here: https://github.com/vmware/hillview/pull/650#discussion_r462205270

web/src/main/webapp/ui/menu.ts
daniarherikurniawan: Yes, because of the border and shadow that add more height

web/src/main/webapp/hillview.css
daniarherikurniawan: I undo this. There is a side effect that cause the first column to expand in width every time I execute `show`

daniarherikurniawan:
There is one problem that I am currently working on. If the user's cursor expanded any expandableMenu, and then it goes up above the first submenu, the submenus are still displayed. I will add a listener above the first submenu and under the last submenu to hide all the menu if the cursor hover that areas.

daniarherikurniawan:
This is the approach I chose to clear the menu. It's hard to tell when is the right time to clear the submenu because the submenu is by default not selected when the cursor expands it. 
image

web/src/main/webapp/hillview.css
mbudiu-vmw: There may be an inner box where you can apply this.

mbudiu-vmw:
Please document this in a comment.
You may also overlay a rectangle with the whole submenu and handle onmouseexit for it.

web/src/main/webapp/hillview.css
daniarherikurniawan: This is not significant for this PR, I am using the original css. Previously, I added the padding because the label is too close to the column border

web/src/main/webapp/ui/menu.ts
```diff
@@ -249,10 +325,91 @@ export class ContextMenu extends BaseMenu<MenuItem> implements IHtmlElement {
...
+        cell.onmouseenter = () => this.selectSubMenu(parentIndex, subIndex);
+        cell.onmouseleave = () => this.selectSubMenu(parentIndex, -1);
+        this.setSubMenuAction(parentIndex, subMenuIndex, mi, enabled);
+
+        // Adding upperBorder: to clear the menu when the cursor hover above 1st submenu
+        if (subMenuPlacementIdx > 0 && this.subMenuCells[parentIndex].cells.length == 1) {
+            const upperBorder = this.rows[subMenuPlacementIdx - 1].insertCell(1);
+            upperBorder.classList.add("dummyMenu");
+            upperBorder.onmouseenter = () => this.clear();
+            this.subMenuCells[parentIndex].upperBorder = upperBorder;
```
daniarherikurniawan, line 391: This is the approach I chose to clear the menu. It's hard to tell when is the right time to clear the submenu because the submenu is by default not selected when the cursor expands it.
image

mbudiu-vmw: test this even if you move the mouse quickly.

web/src/main/webapp/ui/menu.ts
daniarherikurniawan: Addressed this by changing the display from `block` to `table-cell`

web/src/main/webapp/ui/menu.ts
daniarherikurniawan: 1 out of 20 trials, I do catch the bug. I will fix this with a nicer approach. I am thinking of having dummyMenu on the empty spaces above and under the displayed subMenus, these dummyMenu will be transparent and will clear the menus on mouseEnter. 

You mentioned about having rectangle overlay. In my opinion, it is more complex since I need to get the 4 corners locations of the submenu which are dynamically changing.

web/src/main/webapp/ui/menu.ts
```diff
@@ -109,10 +128,27 @@ abstract class BaseMenu<MI extends BaseMenuItem> implements IHtmlElement {
...
     public addItem(mi: MI, enabled: boolean): HTMLTableDataCellElement {
         const index = this.items.length;
         this.items.push(mi);
         const trow = this.tableBody.insertRow();
+
+        const dummySubMenu = trow.insertCell(0);
+        dummySubMenu.classList.add("dummyMenu");
+        dummySubMenu.onmouseenter = () => this.hide();
+        this.subMenuCells[index] = initSubMenu();
+        this.subMenuCells[index].dummySubMenu = dummySubMenu;
```
daniarherikurniawan, line 149: I added dummySubMenus as mouseExit listener in the second column. This simplify the implementation and more robust than the previous approach.
image

PR #651, opened 2020-07-27
Improved instructions for cluster installation

mbudiu-vmw:


mbudiu-vmw:
@daniarherikurniawan : hopefully these instructions are enough to install on a cluster

README.md
```diff
@@ -179,6 +179,14 @@ machine as a one-machine cluster.
 
 ## 3.2 Deployment scripts
 
+First install Hillview locally:
+
+```
+$ bin/install.sh
+```
+
+Edit the config.json file as described above.
```
daniarherikurniawan, line 188: After this, the next instruction is `deploy.py`, but there is no `target/` directories to deploy. I believe there should be 1 more step which is 'bin/rebuild.sh'.  When you wrote `First install Hillview locally:`, it means that we should clone this master repo, right?

mbudiu-vmw: If you didn't clone the repository you don't have this file either.

daniarherikurniawan: Got it! In my case, I still need to run `bin/rebuild.sh` before deploying it

mbudiu-vmw: what happens if you don't?

PR #662, opened 2020-08-06
Added border on static menu and Modified Cassandra manual

daniarherikurniawan:
I added the original menu border, and I also added `border:0` at dropdown menu's css to remove the stretched border above and under the submenu. There is still no border at the menu-block which has an expandable menu.

I also fixed the Cassandra manual to be more precise about the host info.

image
image

### DIFF
--- a/docs/userManual.md
+++ b/docs/userManual.md
@@ -465,9 +465,9 @@ The following menu allows the user to specify the data to load.
   to, which is Cassandra. We also support another DB, 'mysql' and 
   'impala'.
 
-* host: The network name/IP address of Hillview's root node. A root 
-  node is responsible for running Hillview web server interface as  
-  well as Cassandra instance at the background.
+* host: One of the network name/IP address of the machines hosting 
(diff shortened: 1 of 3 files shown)


docs/userManual.md
```diff
@@ -465,9 +465,9 @@ The following menu allows the user to specify the data to load.
   to, which is Cassandra. We also support another DB, 'mysql' and 
   'impala'.
 
-* host: One of the network name/IP address of the machines hosting 
-  Cassandra cluster. Every node in Cassandra cluster are the same 
-  (no master-slave concept), thus Hillview can connect to any node.
+* host: The network name/IP address of Hillview's root node. A root
```
mbudiu-vmw, line 468: are you sure about this?
what is this address used for?
if this is correct you have to also say in the deployment section that the root node has to be colocated with a cassandra node.

daniarherikurniawan: Sorry, my mistake. I just did more testing and now Hillview can connect to any cassandra node. The original description should be fine. I will revert the file.

PR #663, opened 2020-08-10
Adding Bookmark Support

daniarherikurniawan:
This PR will enable bookmark feature. 
1. Client will get a copyable link (such as: localhost:8080/bookmark?id=9f528817-fd38-4895-91bf-43feb4e4e4a7) when the bookmark button is clicked.   
2. The link will trigger an http GET request at `bookmark/`, thus I  added a servlet to handle `bookmark/` and extract the `id=`. 
3. The servlet will store the `bookmarkID` to the cookie, and then redirect the client to `localhost:8080/`
4. The bookmark will be loaded based on to the `bookmarkID` cookie
5. The cookie will be deleted once it is read

Currently, the code will create a `bookmark/` folder if it doesn't exist. Perhaps we want to create this folder by default during installation. 

image


web/src/main/webapp/loadMenu.ts
```diff
@@ -49,6 +49,8 @@ export class LoadMenu extends RemoteObject implements IDataView {
         this.console = new ErrorDisplay();
         this.top.appendChild(this.console.getHTMLRepresentation());
         this.getUIConfig();
+        // Check whether the user is trying to visit a bookmark link
+        this.tryOpeningBookmark();
```
mbudiu-vmw, line 53: use the argument instead as described above.

web/src/main/webapp/loadMenu.ts
```diff
@@ -747,3 +760,31 @@ class PingReceiver extends OnCompleteReceiver<string[]> {
         this.page.reportError(value.toString());
     }
 }
+
+class CreateBookmarkContentReceiver extends Receiver<string> {
```
mbudiu-vmw, line 764: OnCompleteReceiver.

web/src/main/webapp/ui/dialog.ts
```diff
@@ -740,6 +740,39 @@ export class NotifyDialog extends DialogBase {
...
+            .on("start", () => this.dragEnd());
+
+        const urlBox = makeInputBox(url) as HTMLInputElement;
+        urlBox.style.width = "-webkit-fill-available";
+        this.fieldsDiv.appendChild(urlBox);
+
+        const copyButton = document.createElement("button");
+        copyButton.textContent = "Copy Link";
+        copyButton.classList.add("confirm");
+        copyButton.addEventListener("click", () => {
```
mbudiu-vmw, line 761: nice

web/src/main/webapp/ui/dialog.ts
```diff
@@ -740,6 +740,39 @@ export class NotifyDialog extends DialogBase {
...
+ */
+export class NotifyBookmarkDialog extends NotifyDialog {
+    constructor(title: string, url: string | null, toolTip: string) {
+        super(title, null, toolTip);
+
+        const nodrag = d3drag()
+            .on("start", () => this.dragEnd());
+
+        const urlBox = makeInputBox(url) as HTMLInputElement;
+        urlBox.style.width = "-webkit-fill-available";
```
mbudiu-vmw, line 755: webkit? In general we try to write portable code across all browsers. Is there a portable version of this attribute?

web/src/main/java/org/hillview/BookmarkServlet.java
```diff
@@ -0,0 +1,49 @@
...
+import javax.servlet.http.*;
+import org.apache.commons.io.FileUtils;
+
+@WebServlet("/bookmark")
+public class BookmarkServlet extends HttpServlet {
+    private static final long serialVersionUID = 1L;
+    public static final String bookmarkDirectory = "bookmark";
+    public static final String bookmarkExtension = ".txt";
+
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
```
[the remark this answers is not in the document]
daniarherikurniawan: This doesn't work as expected because HillviewToplevel's constructor will be initialized before index.html calling `onload()`. Specifically, HillviewToplevel constructor is initialized  when hillview.ts does `import {createHillview} from "./toplevel";`. 

I will use the HillviewToplevel to pass the bookmark filename, but not as the constructor argument. Then, I will delete the servlet because now the index.html can handle input params from the url. The bookmark will looks like this: `http://localhost:8080/?bookmark=d15d96d5-ae0b-407e-bc84-f0f2b98812ab.txt`

web/src/main/webapp/hillview.ts
```diff
@@ -15,6 +15,11 @@
  * limitations under the License.
  */
 
+// Check whether there is a bookmark parameter in the url
+const url = new URL(window.location.href);
```
mbudiu-vmw, line 19: this is an interesting technique. I like it.

web/src/main/java/org/hillview/targets/InitialObjectTarget.java
```diff
@@ -50,6 +54,9 @@
     private static final String LOCALHOST = "127.0.0.1";
     private static final String ENV_VARIABLE = "WEB_CLUSTER_DESCRIPTOR";
 
+    public static final String bookmarkDirectory = "bookmark";
+    public static final String bookmarkExtension = ".txt";
```
mbudiu-vmw, line 58: perhaps it should be json

web/src/main/webapp/loadMenu.ts
```diff
@@ -42,13 +42,15 @@ export class LoadMenu extends RemoteObject implements IDataView {
...
 
-    constructor(protected init: InitialObject, protected page: FullPage) {
+    constructor(protected init: InitialObject, protected page: FullPage, protected bookmarkFile: string) {
         super(init.remoteObjectId);
         this.viewKind = "Load";
         this.top = document.createElement("div");
         this.console = new ErrorDisplay();
         this.top.appendChild(this.console.getHTMLRepresentation());
         this.getUIConfig();
+        // Check whether the user is trying to visit a bookmark link
```
mbudiu-vmw, line 52: this is actually a small race, because both getUIConfig and openingBookmark will execute asynchronously, but it will probably work.

PR #672, opened 2020-08-17
Check for null cells in cassandra tables

mbudiu-vmw:
@daniarherikurniawan : without this check the `testCassandraTypeConversion` test started failing on my machine. Perhaps Cassandra optimized the storage after a while?


platform/src/main/java/org/hillview/storage/CassandraSSTableLoader.java
```diff
@@ -371,82 +371,88 @@ private static void loadColumns(SSTableReader ssTableReader, List<IAppendableCol
...
-                                        }
-                                        break;
-                                    case EMPTY:
-                                        break;
-                                }
+                                col.appendMissing();
+                                currentColumn++;
+                                continue;
+                            }
+                            Object value = arrSerializers[currentColumn].deserialize(byteBuff);
```
mbudiu-vmw, line 386: there are no changes at all in this section, just indentation.

PR #675, opened 2020-08-19
Fix the problem on loading Federated Menu

daniarherikurniawan:
Originally, the federated menu won't load, due this logic:
```
const connection = dialog.getCassandraConnection();
if (connection == null) {
    this.page.reportError("Not a valid connection");
    return;
}
```
The `dialog.getCassandraConnection()` will always return `null`, thus it displays  `Not a valid connection`. This `getCassandraConnection()` will be executed when the user presses the `confirm` button. Therefore, I remove the null checking, and make it similar to ` getJdbcConnection()`.

In the next PR, I will add Hive connection dialog on top of this logic.

### DIFF
--- a/web/src/main/webapp/initialObject.ts
+++ b/web/src/main/webapp/initialObject.ts
@@ -218,7 +218,8 @@ export class InitialObject extends RemoteObject {
         rr.invoke(observer);
     }
 
-    public loadFederatedDBTable(conn: any, db: String, loadMenuPage: FullPage): void {
+    public loadFederatedDBTable(conn: JdbcConnectionInformation | CassandraConnectionInfo,
+            db: String, loadMenuPage: FullPage): void {
         switch (db) {
             case "mysql":
             case "impala":
--- a/web/src/main/webapp/initialObject.ts
+++ b/web/src/main/webapp/initialObject.ts
@@ -218,11 +218,15 @@ export class InitialObject extends RemoteObject {
         rr.invoke(observer);
     }
 
-    public loadFederatedDBTable(jdbcConn: JdbcConnectionInformation, cassConn: CassandraConnectionInfo, loadMenuPage: FullPage): void {
-        if (jdbcConn.databaseKind == "cassandra") {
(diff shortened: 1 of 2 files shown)


daniarherikurniawan:
I will create another PR that will include Hive menu

mbudiu-vmw:
So is this PR closed by mistake?

daniarherikurniawan:
Not by mistake. Currently, I am refactoring the DBDialog. I will fix the
issue and add new hive menu in the incoming PR.

mbudiu-vmw:
In general a fix should be in a separate PR from a new feature.

daniarherikurniawan:
Okay, please review this code

web/src/main/webapp/loadMenu.ts
```diff
@@ -238,12 +238,8 @@ export class LoadMenu extends RemoteObject implements IDataView {
                 text: "Federated DB tables...",
                 action: () => {
                     const dialog = new DBDialog(true);
-                    const connection = dialog.getCassandraConnection();
-                    if (connection == null) {
-                        this.page.reportError("Not a valid connection");
-                        return;
-                    }
-                    dialog.setAction(() => this.init.loadFederatedDBTable(dialog.getConnection(), connection, this.page));
+                    dialog.setAction(() => this.init.loadFederatedDBTable(dialog.getJdbcConnection(),
```
mbudiu-vmw, line 241: how about instead checking if this is cassandra and then extracting only the proper connection?
Probably loadFederatedDBTable should have a single argument, not two.

daniarherikurniawan: I updated this code. The connection object will be passed as an `any` object. Thus, it can handle multiple connection types including HiveConnection without adding more input argument.

mbudiu-vmw:
Try to avoid using `any`. Typescript has union types, so you write `number | string`, for example.

daniarherikurniawan:
Updated with the union types. Thanks!

PR #684, opened 2020-09-02
Add sub menu container in the left side of main menu

daniarherikurniawan:
This PR will handle the case where the submenu must be displayed on the left side of the main menu. It is done by adding an empty submenu container on the left of the main menu. When there is not enough room to display the submenu on the right, the submenu will be placed into the container. The container will copy all attributes of the current submenus that need to be displayed. When hiding the submenu, all attributes will be reset and the user won't be able to notice that there are empty containers on the left of the main menu.

### DIFF
--- a/web/src/main/webapp/ui/menu.ts
+++ b/web/src/main/webapp/ui/menu.ts
@@ -133,7 +133,7 @@ abstract class BaseMenu<MI extends BaseMenuItem> implements IHtmlElement {
     }
 
     public hide(): void {
-        // this.outer.classList.add("hidden");
+        this.outer.classList.add("hidden");
     }
 
     public getCell(mi: MI): HTMLTableCellElement {
--- a/web/src/main/webapp/ui/menu.ts
+++ b/web/src/main/webapp/ui/menu.ts
@@ -61,16 +61,28 @@ abstract class BaseMenu<MI extends BaseMenuItem> implements IHtmlElement {
     public outer: HTMLTableElement;
     public tableBody: HTMLTableSectionElement;
     public cells: HTMLTableDataCellElement[];
+    /** 
+     * All submenus are stored here, and normally will be displayed 
+     * on the right side of the main menu. If there is not enough space,
+     * the submenus attributes will be copied to leftSubMenuCells to be 
+     * displayed on the left side of main menu.
+    */
     public subMenuCells: SubMenuCells[];
+    // This will hold the submenu cells that are displayed on the left side
+    public leftSubMenuCells: HTMLTableDataCellElement[];
     public rows: HTMLTableRowElement[];
     public selectedIndex: number; // -1 if no item is selected
     public selectedParentMenu: number; // -1 if no item is selected
     public selectedSubMenu: number; // -1 if no item is selected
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
I will test this. It looks a bit complicated; I was hoping there's a way to do this using some CSS tricks.

daniarherikurniawan:
The complication is only about resetting the CSS of the left container after displaying the submenus. We display the menu in a columnar format, so it's a bit hard to rely on CSS tricks when we need to shift the whole column.

web/src/main/webapp/ui/menu.ts
```diff
@@ -61,23 +61,38 @@ abstract class BaseMenu<MI extends BaseMenuItem> implements IHtmlElement {
...
+        this.leftSubMenuCells = [];
         this.selectedParentMenu = -1;
         this.rows = [];
         this.selectedIndex = -1;
         this.outer = document.createElement("table");
         this.outer.classList.add("menu", "hidden");
         this.tableBody = this.outer.createTBody();
         this.outer.onkeydown = (e) => this.keyAction(e);
+        this.longestSubMenuText = 0;
+        this.pixelPerChar = 8;
```
mbudiu-vmw, line 94: you cannot make this assumption.

web/src/main/webapp/ui/menu.ts
```diff
@@ -134,9 +149,25 @@ abstract class BaseMenu<MI extends BaseMenuItem> implements IHtmlElement {
...
         this.selectedParentMenu = index;
         if (index < 0) throw new Error("Cannot find menu item " + parentMenu);
+        let i = index;
         this.subMenuCells[index].cells.forEach((cell) => {
-            cell.style.display = "table-cell";
-            cell.classList.add("menuShadow");
+            if (this.subMenuOnLeftSide) {
+                // this.leftSubMenuCells[i].textContent = cell.textContent;
+                this.leftSubMenuCells[i].innerHTML = cell.innerHTML;
+                const leftSubMenu = this.leftSubMenuCells[i];
```
mbudiu-vmw, line 157: why not move this one line up

web/src/main/webapp/ui/menu.ts
```diff
@@ -318,7 +378,20 @@ export class ContextMenu extends BaseMenu<MenuItem> implements IHtmlElement {
             y = max.height - this.outer.offsetHeight - 5;
         if (y < 0)
             y = 0;
-        this.move(x, y);
+
+        this.subMenuOffsetWidth = this.longestSubMenuText * this.pixelPerChar;
```
mbudiu-vmw, line 382: what you should do is really measure the width of the object as rendered on the screen, do not attempt to compute the size in this simple way.

mbudiu-vmw (changes requested):

web/src/main/webapp/ui/menu.ts
daniarherikurniawan: We can't get the width because the submenu (the cell) is not yet displayed. Now, I approximate the width based on the main menu since the main menu will be displayed and has `offsetWidth` value.

PR #686, opened 2020-09-03
Read from Cassandra's snapshot

daniarherikurniawan:
This PR contains the backend code for loading Cassandra Snapshot. The UI is not yet updated, because the code depends on the other PR (https://github.com/vmware/hillview/pull/675). 

I added some codes about discovering the snapshot directory and validating the snapshot name + snapshot folders.  
The (platform) unit test should be working as long as you rerun `loadCassandraDB.sh` which will create the snapshot with this command `${CASSANDRA_INSTALLATION_DIR}/bin/nodetool snapshot myTestSnapshot cassdb`

### DIFF
--- a/platform/src/test/java/org/hillview/test/storage/SSTableTest.java
+++ b/platform/src/test/java/org/hillview/test/storage/SSTableTest.java
@@ -54,6 +54,7 @@ public class SSTableTest extends BaseTest {
         conn.user = "";
         conn.password = "";
         conn.lazyLoading = true;
+        conn.snapshotName = "myTestSnapshot";
         return conn;
     }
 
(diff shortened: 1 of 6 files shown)


platform/src/test/java/org/hillview/test/storage/SSTableTest.java
```diff
@@ -54,6 +54,7 @@ public CassandraConnectionInfo getConnectionInfo() {
         conn.user = "";
         conn.password = "";
         conn.lazyLoading = true;
+        conn.snapshotName = "myTestSnapshot";
```
mbudiu-vmw, line 57: Do you have to add this to the UI as well?

daniarherikurniawan: Yes, I will do some changes in the UI and let you know when it's ready

daniarherikurniawan:
The UI menu is updated, you can insert `myTestSnapshot` as the snapshot name to test it out. The snapshot name for testing is specified at loadCassandraDB.sh.

PR #718, opened 2021-06-09
Remove erroneous pip install command

wkk:
`json` is a built-in python module and can't to be installed via `pip`.

### DIFF
--- a/bin/install-dependencies.sh
+++ b/bin/install-dependencies.sh
@@ -30,7 +30,6 @@ esac
 ${SUDO} ${INSTALL} install wget maven ${NODEJS} ${LIBFORTRAN} unzip gzip python3
 echo "Installing typescript compiler"
 ${SUDO} npm install -g typescript@3.9.7
-pip install json
 
 # Download apache if not there.
 pushd ..


mbudiu-vmw (commented):
Why is this wrong?
This seems to be needed only by the deploy-greenplum.py script.

wkk:
1. `json` is a built-in module of python. After installing python you can just `import json` directly.
2. When you run `pip install json`, `pip` will look for a package named `json` on PyPI, which will cause an error because there is no such package. Below is the error I got on Ubuntu:
    ```shell
    $ pip install json
    ERROR: Could not find a version that satisfies the requirement json (from versions: none)
    ERROR: No matching distribution found for json
    ```

mbudiu-vmw:
Interesting. These scripts were written to work with both python2 and python3; maybe this is a python2 module.

mbudiu-vmw:
@wkk: please see the instructions from the bot about the CLA; it's fairly standard.

wkk:
@mbudiu-vmw done

mbudiu-vmw (approved):

PR #720, opened 2021-06-29
Fix error when building and testing on Ubuntu 20.04

wkk:
* Update `jblas` to 1.2.5
  The 1.2.4 version doesn't work with `libgfortran5` on ubuntu 20.04.

* Use `-no-check-certificate` option with `wget` when downloading the on time data
  Without this option `wget` will fail with the error message "Unable to locally verify the issuer's authority."

### DIFF
--- a/data/ontime/download.py
+++ b/data/ontime/download.py
@@ -6,7 +6,8 @@ import os
 import subprocess
 from sys import platform
 
-wgetoptions = ""
+wgetoptions = "--no-check-certificate"
+
 
 def execute_command(command):
     """Executes the specified command using a shell"""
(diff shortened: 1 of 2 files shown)


mbudiu-vmw:
I assume this works with older versions of Ubuntu too?

data/ontime/download.py
```diff
@@ -45,17 +48,6 @@ def download(startyear, startmonth, endyear, endmonth):
         else:
             print("Already having " + basename)
 
-def main():
```
mbudiu-vmw, line 48: are you sure this code is no longer needed?

wkk: Yes. Basically this part adds `--cipher 'DEFAULT:!DH'` option to `wget` if it detects the system is ubuntu20.04. With the `--no-check-certificate` option I don't think this is needed anymore.

wkk: > I assume this works with older versions of Ubuntu too?

Yes, they should work with ubuntu 18.04 as well.

There is another fix for the wget problem though, which is to check in the [entrust root certificate](https://web.entrust.com/root-certificates/entrust_l1k.cer) to our repo, and then replace the `--no-check-certificate` option with `--ca-certificate=entrust_l1k.cer` so it will use the provided certificate. It might be safer that way?

wkk (commented):
> I assume this works with older versions of Ubuntu too?

Yes, they should work with ubuntu 18.04 as well.

There is another fix for the wget problem though, which is to check in the [entrust root certificate](https://web.entrust.com/root-certificates/entrust_l1k.cer) to our repo, and then replace the `--no-check-certificate` option with `--ca-certificate=entrust_l1k.cer` so that `wget` will use the provided certificate. It might be safer that way?

mbudiu-vmw:
Does this make the CI work?

wkk:
> Does this make the CI work?

Unfortunately no. In GitHub actions `wget` fails to download even with this fix. I think it may be because the azure IP address used for GitHub actions is blocked by bts.gov.

PR #722, opened 2021-07-14
Implement reading from delta table

wkk:
* Implemented reading from delta table (both backend and frontend)
* Added a script for generating a test delta table (by adding a `-g` option to `demo-data-cleaner.sh`). Currently the dependencies are not installed by default.
* Added a test case for reading from delta table.

### DIFF
--- a/platform/pom.xml
+++ b/platform/pom.xml
@@ -160,16 +160,6 @@
             <version>7.1</version>
         </dependency>
         <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-hdfs-client</artifactId>
-            <version>3.3.0</version>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
-            <version>2.10.5.1</version>
-        </dependency>
-        <dependency>
             <groupId>commons-pool</groupId>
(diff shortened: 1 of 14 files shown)


mbudiu-vmw (commented):
This looks pretty good, but I will first try to run it.

platform/pom.xml
```diff
@@ -159,16 +159,6 @@
             <artifactId>asm</artifactId>
             <version>7.1</version>
         </dependency>
-        <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-hdfs-client</artifactId>
-            <version>3.3.0</version>
-        </dependency>
```
wkk, line 166: Removed because it's a dependency of `hadoop-client`

platform/pom.xml
```diff
@@ -159,16 +159,6 @@
...
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
-            <version>2.10.5.1</version>
-        </dependency>
         <dependency>
```
wkk, line 162: Removed because delta standalone reader isn't compatible with 2.10.5.1: it uses an older version ( 2.8.4) of `jackson-databind` instead. Not sure if this is the right thing to do.

wkk:
Closing for now. Will make a new PR later

PR #725, opened 2021-07-23
Implement reading from delta table

wkk:
* Implemented reading from delta table (both backend and frontend)
* Added a script for generating a test delta table (by adding a `-g` option to `demo-data-cleaner.sh`). Currently the dependencies are not installed by default.
* Added a test case for reading from delta table.

### DIFF
--- a/bin/demo-data-cleaner.sh
+++ b/bin/demo-data-cleaner.sh
@@ -5,9 +5,24 @@
 set -e
 
 export MAVEN_OPTS="-Xmx2048M"
+
 pushd ../data/ontime
 ./download.py
 popd
+
+while getopts "g" opt; do
+  case "$opt" in
+    g)
+      python3 generate-delta-table.py
+      ;;
+    *)
+      echo "demo-data-cleaner.sh [-g]"
+      echo "-g: Generating delta table for testing, requires python3, pyspark, and delta-spark"
+      exit 1
+      ;;
+  esac
+done
+
 pushd ../platform
 java -jar target/data-cleaner-jar-with-dependencies.jar
 popd
--- a/bin/demo-data-cleaner.sh
+++ b/bin/demo-data-cleaner.sh
@@ -17,7 +17,8 @@ while getopts "g" opt; do
       ;;
     *)
       echo "demo-data-cleaner.sh [-g]"
-      echo "-g: Generating delta table for testing, requires python3, pyspark, and delta-spark"
+      echo "This script runs the java class DemoDataCleaner to prepare data for testing/demo purposes"
+      echo "-g: Also generate a delta table for testing, requires python3, pyspark, and delta-spark"
       exit 1
       ;;
   esac
--- a/platform/pom.xml
+++ b/platform/pom.xml
@@ -160,16 +160,6 @@
             <version>7.1</version>
         </dependency>
         <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-hdfs-client</artifactId>
-            <version>3.3.0</version>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
-            <version>2.10.5.1</version>
-        </dependency>
-        <dependency>
             <groupId>commons-pool</groupId>
             <artifactId>commons-pool</artifactId>
             <version>1.6</version>
@@ -188,7 +178,7 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
-            <version>3.3.0</version>
+            <version>3.3.1</version>
             <exclusions>
                 <exclusion>
                     <groupId>com.nimbusds</groupId>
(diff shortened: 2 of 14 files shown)


web/src/main/java/org/hillview/targets/InitialObjectTarget.java
```diff
@@ -88,6 +93,8 @@ public InitialObjectTarget() {
 
     private void initialize(final HostList description) {
         this.emptyDataset = RemoteDataSet.createCluster(description, RemoteDataSet.defaultDatasetIndex);
+        PingSketch<Empty> ping = new PingSketch<Empty>();
+        this.workers = new ArrayList<>(Objects.requireNonNull(this.emptyDataset.blockingSketch(ping)));
```
mbudiu-vmw, line 97: we have our own function checkNonNull. Perhaps we shouldn't have, but let's use it.

platform/src/main/java/org/hillview/storage/delta/DeltaTableDescription.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * This class contains necessary information to access a delta table
+ */
+public class DeltaTableDescription implements Serializable {
+    static final long serialVersionUID = 1;
+
+    public String path;
```
mbudiu-vmw, line 17: Can you also add a toString method?

platform/src/main/java/org/hillview/maps/FindDeltaTableFilesMap.java
```diff
@@ -0,0 +1,41 @@
...
+import org.hillview.utils.HillviewLogger;
+import org.hillview.utils.Utilities;
+
+import javax.annotation.Nullable;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class FindDeltaTableFilesMap<T> implements IMap<T, List<IFileReference>> {
```
mbudiu-vmw, line 15: Let's rename this class into ExtractWorkerFilesMap. It does not really find anything.

web/src/main/webapp/initialObject.ts
```diff
@@ -48,7 +53,7 @@ export interface Merged {
     second: DataLoaded;
 }
 
-export type DataLoaded = FilesLoaded | TablesLoaded | HillviewLogs | IDatasetSerialization | Merged;
+export type DataLoaded = FilesLoaded | TablesLoaded | DeltaTableLoaded | HillviewLogs | IDatasetSerialization | Merged;
```
mbudiu-vmw, line 56: Do you need here a new type or can you just reuse FilesLoaded?

mbudiu-vmw (approved):
I will approve, but perhaps you can address the few comments before we merge.

bin/demo-data-cleaner.sh
```diff
@@ -5,9 +5,24 @@
...
 popd
+
+while getopts "g" opt; do
+  case "$opt" in
+    g)
+      python3 generate-delta-table.py
+      ;;
+    *)
+      echo "demo-data-cleaner.sh [-g]"
+      echo "-g: Generating delta table for testing, requires python3, pyspark, and delta-spark"
```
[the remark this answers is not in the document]
wkk: Updated usage description

web/src/main/java/org/hillview/targets/InitialObjectTarget.java
wkk: I think it's better to use the one provided by java, but I have changed to our checkNull function for now

platform/src/main/java/org/hillview/storage/delta/DeltaTableDescription.java
wkk: Done

platform/src/main/java/org/hillview/maps/FindDeltaTableFilesMap.java
wkk: Done

web/src/main/webapp/initialObject.ts
wkk: I think we should have a new type due to the following reasons:
1. `FilesLoaded` needs a `FileSetDescription`, current that is generated at the back end and not returned to the front end.
2. Having a separate `DeltaTableLoaded` interface allows the front end to update the UI specifically for delta table. For example, current we can set the page title as "loaded from delta table" rather than "loaded from files", otherwise it can be confusing for the user. In the future we can also add a drop down menu for user to choose snapshot version once a delta table is loaded.

platform/pom.xml
```diff
@@ -159,16 +159,6 @@
             <artifactId>asm</artifactId>
             <version>7.1</version>
         </dependency>
-        <dependency>
```
mbudiu-vmw, line 162: are you sure these can be removed? they are here to prevent conflicts, they are transitive dependencies of other dependencies we use.

mbudiu-vmw (approved):

mbudiu-vmw:
I will merge this, but we should consider the conflicts in the pom file.
Perhaps leave this as a low priority item.

PR #727, opened 2021-07-29
Implement writing table to parquet

### LINKED ISSUE #346: Add a parquet test file to the repository.

wkk:
* Implement `ParquetFileWriter` that can write an `ITable` to a `.parquet` file.
* Generate test parquet files in `DemoDataCleaner` (closes #346).
* Modified `ParquetTest` to use the generated parquet file.

### DIFF
--- a/bin/demo-data-cleaner.sh
+++ b/bin/demo-data-cleaner.sh
@@ -10,20 +10,25 @@ pushd ../data/ontime
 ./download.py
 popd
 
-while getopts "g" opt; do
+ARGS=""
+while getopts "dp" opt; do
   case "$opt" in
-    g)
+    d)
       python3 generate-delta-table.py
       ;;
+    p)
+      ARGS+="-Dparquet.enabled "
+      ;;
     *)
-      echo "demo-data-cleaner.sh [-g]"
-      echo "This script runs the java class DemoDataCleaner to prepare data for testing/demo purposes"
-      echo "-g: Also generate a delta table for testing, requires python3, pyspark, and delta-spark"
+      echo "Usage: demo-data-cleaner.sh [-g]"
+      echo "This script runs the java class DemoDataCleaner to prepare data for testing/demo purposes."
+      echo "-d: Also generate a delta table for testing, requires python3, pyspark, and delta-spark."
+      echo "-p: Also generate parquet files for testing."
       exit 1
       ;;
   esac
 done
 
 pushd ../platform
-java -jar target/data-cleaner-jar-with-dependencies.jar
+java -jar ${ARGS} target/data-cleaner-jar-with-dependencies.jar
 popd
--- a/bin/demo-data-cleaner.sh
+++ b/bin/demo-data-cleaner.sh
@@ -20,7 +20,7 @@ while getopts "dp" opt; do
       ARGS+="-Dparquet.enabled "
       ;;
     *)
-      echo "Usage: demo-data-cleaner.sh [-g]"
+      echo "Usage: demo-data-cleaner.sh [-d] [-p]"
       echo "This script runs the java class DemoDataCleaner to prepare data for testing/demo purposes."
       echo "-d: Also generate a delta table for testing, requires python3, pyspark, and delta-spark."
       echo "-p: Also generate parquet files for testing."
--- a/platform/src/main/java/org/hillview/main/DemoDataCleaner.java
+++ b/platform/src/main/java/org/hillview/main/DemoDataCleaner.java
@@ -17,6 +17,7 @@
 
 package org.hillview.main;
 
+import org.hillview.storage.ParquetFileWriter;
 import org.hillview.table.LazySchema;
 import org.hillview.storage.CsvFileLoader;
 import org.hillview.storage.CsvFileWriter;
@@ -83,6 +84,27 @@ class DemoDataCleaner {
                         OrcFileWriter owriter = new OrcFileWriter(end);
                         System.out.println("Writing " + end);
                         owriter.writeTable(p);
+                    }
+
+                    final String parquetFileName = end.replace(".orc", ".parquet");
+                    File parquetFile = new File(parquetFileName);
+                    if (!parquetFile.exists()) {
+                        ParquetFileWriter writer = new ParquetFileWriter(parquetFileName);
+                        System.out.println("Writing " + parquetFileName);
+                        try {
+                            writer.writeTable(p);
+                        } catch (RuntimeException runtimeException) {
+                            System.err.println("Error when writing to parquet file: " + runtimeException.getMessage());
+                            // If the exception happens during writing, an incomplete file may be left
+                            try {
+                                Files.deleteIfExists(parquetFile.toPath());
+                            } catch (IOException ioException) {
+                                System.err.println("Auto Deletion failed: " + ioException.getMessage());
+                                System.err.println("Please manually delete " + parquetFile.getPath());
+                                System.exit(-1);
+                            }
+                        }
+
                     }
 
                     String big = filename.replace(".csv.gz", ".orc");
--- a/platform/src/main/java/org/hillview/main/DemoDataCleaner.java
+++ b/platform/src/main/java/org/hillview/main/DemoDataCleaner.java
@@ -43,6 +43,8 @@ class DemoDataCleaner {
     private static final String dataFolder = "../data/ontime";
 
     public static void main(String[] args) throws IOException {
+        boolean parquetEnabled = (System.getProperty("parquet.enabled") != null);
+
         HillviewLogger.initialize("data cleaner", "hillview.log");
         String prefix = "On_Time_On_Time_Performance_";
         Path folder = Paths.get(dataFolder);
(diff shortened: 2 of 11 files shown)


platform/src/main/java/org/hillview/main/DemoDataCleaner.java
```diff
@@ -85,6 +86,27 @@ public static void main(String[] args) throws IOException {
                         owriter.writeTable(p);
                     }
 
+                    final String parquetFileName = end.replace(".orc", ".parquet");
```
mbudiu-vmw, line 89: Can you make this conditional on a command -line flag?

platform/src/main/java/org/hillview/storage/ParquetFileWriter.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public ParquetFileWriter(String filepath) {
+        this.filepath = filepath;
+    }
+
+    private static MessageType getMessageType(Schema schema) {
+        Types.MessageTypeBuilder builder = Types.buildMessage();
+
+        for (ColumnDescription description : schema.getColumnDescriptions()) {
+            switch (description.kind) {
```
mbudiu-vmw, line 33: I think that we should use native parquet types when they exist. E.g. Dates. Otherwise loading this data into an external dB will make no sense. What does the orc writer do?

mbudiu-vmw (changes requested):

mbudiu-vmw:
Can you also write a few more tests? E.g. Write and read back give same result?

platform/src/main/java/org/hillview/main/DemoDataCleaner.java
wkk: Yes, I'll add a flag

platform/src/main/java/org/hillview/storage/ParquetFileWriter.java
wkk: The orc writer does try to use native types if exists. I'll try to do that as well.

wkk:
> Can you also write a few more tests? E.g. Write and read back give same result?

Sure. In that case may also need to make some changed to the `ParquetFileReader`

platform/src/main/java/org/hillview/main/DemoDataCleaner.java
wkk: Added a flag `-p` to `demo-data-cleaner.sh` to control the generation of parquet files. Also changed the flag for delta table from `-g` to `-d`.

platform/src/main/java/org/hillview/storage/ParquetFileWriter.java
wkk: Added native logical type support. Also added a new test that checks reading back after write gives the same result

platform/src/main/java/org/hillview/storage/ParquetFileLoader.java
```diff
@@ -127,8 +173,23 @@ private static void appendGroup(
...
-                    String s = g.getString(field, 0);
-                    col.append(s);
+                    if (logicalType instanceof IntervalLogicalTypeAnnotation) {
+                        ByteBuffer bb = g.getBinary(field, 0).toByteBuffer();
+                        bb.order(ByteOrder.LITTLE_ENDIAN);
+                        int months = bb.getInt();
+                        int days = bb.getInt();
+                        int milliseconds = bb.getInt();
+                        if (months != 0) {
+                            HillviewLogger.instance.warn("Month is not zero in parquet interval: using conversion of 30 days per month",
```
mbudiu-vmw, line 183: this may trigger millions of warnings when reading a data file. Perhaps log only once.

platform/src/main/java/org/hillview/storage/ParquetFileLoader.java
```diff
@@ -127,8 +173,23 @@ private static void appendGroup(
...
+                                    "name: {0}, months: {1}, days: {2}, milliseconds: {3}", cds.get(field).getPath(), months, days, milliseconds);
+                        }
+                        int daysInMonth = 30;
+                        double totalMilliseconds = (days + daysInMonth * months) * Converters.SECONDS_TO_DAY * Converters.MILLIS_TO_SECONDS + milliseconds;
```
mbudiu-vmw, line 187: please fold very long lines

platform/src/main/java/org/hillview/storage/ParquetFileLoader.java
```diff
@@ -127,8 +173,23 @@ private static void appendGroup(
...
+                        ByteBuffer bb = g.getBinary(field, 0).toByteBuffer();
+                        bb.order(ByteOrder.LITTLE_ENDIAN);
+                        int months = bb.getInt();
+                        int days = bb.getInt();
+                        int milliseconds = bb.getInt();
+                        if (months != 0) {
+                            HillviewLogger.instance.warn("Month is not zero in parquet interval: using conversion of 30 days per month",
+                                    "name: {0}, months: {1}, days: {2}, milliseconds: {3}", cds.get(field).getPath(), months, days, milliseconds);
+                        }
+                        int daysInMonth = 30;
```
mbudiu-vmw, line 186: is this the formula? it looks weird.

platform/src/main/java/org/hillview/storage/ParquetFileLoader.java
```diff
@@ -153,19 +214,51 @@ private static void appendGroup(
...
+                    kind = ContentsKind.String;
+                }
+                break;
+            case BINARY:
+                if (logicalType instanceof StringLogicalTypeAnnotation) {
+                    kind = ContentsKind.String;
+                } else if (logicalType instanceof JsonLogicalTypeAnnotation) {
+                    kind = ContentsKind.Json;
+                } else {
+                    HillviewLogger.instance.warn("Encountered BINARY field with no or unhandled logical type annotation: attempting to parse as string",
```
mbudiu-vmw, line 258: this will also give lots of warnings

platform/src/main/java/org/hillview/storage/ParquetFileLoader.java
```diff
@@ -127,8 +173,23 @@ private static void appendGroup(
...
+                        int milliseconds = bb.getInt();
+                        if (months != 0) {
+                            HillviewLogger.instance.warn("Month is not zero in parquet interval: using conversion of 30 days per month",
+                                    "name: {0}, months: {1}, days: {2}, milliseconds: {3}", cds.get(field).getPath(), months, days, milliseconds);
+                        }
+                        int daysInMonth = 30;
+                        double totalMilliseconds = (days + daysInMonth * months) * Converters.SECONDS_TO_DAY * Converters.MILLIS_TO_SECONDS + milliseconds;
+                        col.append(totalMilliseconds);
+                    } else {
+                        String s = g.getString(field, 0);
```
mbudiu-vmw, line 190: is this always going to work? 
Is this how parquet encodes strings?

platform/src/main/java/org/hillview/storage/ParquetFileLoader.java
```diff
@@ -263,4 +335,25 @@ public ITable load() {
...
+        public List<? extends IColumn> loadColumns(List<String> names) {
+            FileMetaData fm = ParquetFileLoader.this.metadata.getFileMetaData();
+            MessageType schema = fm.getSchema();
+            List<Type> list = new ArrayList<Type>();
+            for (Type col : schema.getFields()) {
+                String colName = col.getName();
+                if (names.contains(colName))
+                    list.add(col);
+            }
+            assert list.size() > 0;
```
mbudiu-vmw, line 350: So far I never used java assert; is this equivalent to a throw?

mbudiu-vmw (changes requested):

platform/src/main/java/org/hillview/main/DemoDataCleaner.java
```diff
@@ -42,6 +43,8 @@
     private static final String dataFolder = "../data/ontime";
 
     public static void main(String[] args) throws IOException {
+        boolean parquetEnabled = (System.getProperty("parquet.enabled") != null);
```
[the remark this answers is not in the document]
wkk: Like this `java -jar -Dparquet.enabled target/data-cleaner-jar-with-dependencies.jar`.

The `-D` flag can be used to set a system property. It's mentioned in [this documentation](https://docs.oracle.com/javase/8/docs/technotes/tools/unix/java.html): 

> -Dproperty=value
> Sets a system property value. The property variable is a string with no spaces that represents the name of the property. The value variable is a string that represents the value of the property. If value is a string with spaces, then enclose it in quotation marks (for example -Dfoo="foo bar").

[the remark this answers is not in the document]
mbudiu-vmw: I see it now, thanks.

bin/demo-data-cleaner.sh
```diff
@@ -10,20 +10,25 @@ pushd ../data/ontime
...
       python3 generate-delta-table.py
       ;;
+    p)
+      ARGS+="-Dparquet.enabled "
+      ;;
     *)
-      echo "demo-data-cleaner.sh [-g]"
-      echo "This script runs the java class DemoDataCleaner to prepare data for testing/demo purposes"
-      echo "-g: Also generate a delta table for testing, requires python3, pyspark, and delta-spark"
+      echo "Usage: demo-data-cleaner.sh [-g]"
```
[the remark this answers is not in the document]
wkk: updated

platform/src/main/java/org/hillview/storage/ParquetFileLoader.java
wkk: Fixed. Now only log once for each parquet interval column with non-zero month values.

platform/src/main/java/org/hillview/storage/ParquetFileLoader.java
wkk: done

platform/src/main/java/org/hillview/storage/ParquetFileLoader.java
wkk: Actually this function is the conversion from parquet schema to hillview schema. It only happens once and will log at most one warning message for each column

platform/src/main/java/org/hillview/storage/ParquetFileLoader.java
wkk: Added a warning for `BINARY` and `FIXED_LEN_BYTE_ARRAY` columns with missing or unknown logical type annotations. Should help debugging if error does happen during decoding.

mbudiu-vmw (approved):

mbudiu-vmw (approved):

PR #737, opened 2021-08-09
    Autofill HDFS URI in DeltaTableDialog

wkk:
* Autofill HDFS URI in DeltaTableDialog
* Use login shell to run commands on remote hostsso that that hilview processes can pick up environment variables.

### DIFF
--- a/web/src/main/java/org/hillview/targets/InitialObjectTarget.java
+++ b/web/src/main/java/org/hillview/targets/InitialObjectTarget.java
@@ -17,7 +17,10 @@
 
 package org.hillview.targets;
 
+import com.google.gson.JsonPrimitive;
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/hillview/targets/InitialObjectTarget.java
```diff
@@ -106,6 +109,34 @@ public void getUIConfig(RpcRequest request, RpcRequestContext context) {
         this.returnResult(Configuration.instance.getAsJson(), request, context);
     }
 
+    @HillviewRpc
+    public void getHDFSUri(RpcRequest request, RpcRequestContext context) {
```
mbudiu-vmw, line 113: this stuff is always more complicated than you expect.

mbudiu-vmw (approved):

Issue #747, opened 2021-10-18
Python script to create delta table fails


mbudiu-vmw:
@wkk is this something you could take a look at?
This broke our last build.

wkk:
Hi Mihai, sure I'll look into this. At first glance this seems to be causes by the new release of spark 3.2.0. I'll see we need to roll back to 3.1.2

PR #748, opened 2021-10-20
Specify versions for pyspark and delta-spark

bin-wang:
Spark has a new release 3.2.0 and delta lake isn't compatible with it yet. This patch fixes the versions of both packages.


wkk:
Resolves  #747

